PR #42, opened 2020-07-06
A pass on README

olekscode:
Moved Metacello script to the top, added instructions on how to depend on Microdown using baselines, and fixed the code highlighting

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,9 @@ Under writing...
 Similar to markdonw headers are composed of `#` space text on one line.
 The headers can be from 1 to 6
 
-```
+~~~Markdown
 ### Header Level3
-```
+~~~
 There is not support for other forms of declaration. 
 
 ### Anchors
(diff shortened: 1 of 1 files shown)


Ducasse:
Microdown in not markdown!
So why do you change ``` ?

olekscode:
Here are the three changes that I've made to README together with some explanations.
In fact, only Change 3 is important.
Changes 1 and 2 are simply there to make everything more clean and more pretty (but that's subjective)

### Change 1
If you write ` ``` ... ``` ` you get a plain text block. But if you write ` ```Markdown ... ``` ` you get a code highlighted as Markdown.
So here I simply added the code highlighting to make it more pretty and a bit easier to read.

Screenshot 2020-07-07 at 09 21 53

### Change 2
There were nested ` ``` ` inside code blocks that had to be escaped. In general, there are three ways to do this that I know:

1. By indenting the code inside the code block (but then it also renders with this indentation -- check the picture above on the left)
2. By surrounding everything with 4 backticks instead of 3: ` ```` ... ```...``` ... ```` ` (but I think it's hard to count backticks)
3. By surrounding with 3 tildes instead of 3 backticks: `~~~ ... ``` ... ``` ... ~~~` (I find it the most clean)

So I simply changed the solution for escaping ` ``` ` that I find most readable. It's not an important change:

```
~~~
This is equivalent to ``` ... ``` but it allows you to have backticks inside
~~~
```

### Change 3
The instructions on how to install Microdown were located at the very bottom of README. So I moved them up and added some explanations:

> ## How to install it?
> 
> To install `Microdown`, go to the Playground (Ctrl+OW) in your [Pharo](https://pharo.org/) image and execute the following Metacello script (select it and press Do-it button or Ctrl+D):
> 
> ```Smalltalk
> Metacello new
>   baseline: 'Microdown';
>   repository: 'github://pillar-markup/Microdown/src';
>   load.
> ```
> 
> ## How to depend on it?
> 
> If you want to add a dependency on `Microdown` to your project, include the following lines into your baseline method:
> 
> ```Smalltalk
> spec
>   baseline: 'Microdown'
>   with: [ spec repository: 'github://pillar-markup/Microdown/src' ].
> ```

Issue #41, opened 2020-07-06
Outdated baseline. Class comments are not rendering


olekscode:
When I create a new Pharo 9 image and load the baseline of Microdown, the class comments do not render. However, if I also install the baseline of Pillar, everything is OK. So it seems that some necessary packages of Pillar are igrored by Microdown's baseline.

olekscode:
If I install Microdown, class comments are not rendering.
<img width="1440" alt="Screenshot_2020-07-06_at_17 43 02" src="https://user-images.githubusercontent.com/8271691/86740640-1833be80-c037-11ea-9403-5d3ea10e451d.png">

I described this problem in issue #41 

I received error messages that `PRPillarParser` is missing. And since it is part of the `'parser'` group, I added that group to the baseline. After that everything was fine.

<img width="1440" alt="Screenshot 2020-07-07 at 00 35 26" src="https://user-images.githubusercontent.com/8271691/86740268-cee36f00-c036-11ea-871b-5cb62f02802f.png">

Adding `'parser'` to the baseline seemed to have fixed the problem.
But I don't know if this is the best solution.

olekscode:
This was fixed, closing the issue

PR #43, opened 2020-07-06
41 outdated baseline class comments are not rendering

olekscode:
I have added group `'parser'` from Pillar's baseline to the list of groups that will be loaded by the baseline of Microdown.

### Before

```Smalltalk
spec baseline: 'Pillar' with: [ spec 
   loads: #('rich text exporter');
   repository: 'github://pillar-markup/pillar:dev-8/src' ].
```

### After

```Smalltalk
spec baseline: 'Pillar' with: [ spec 
   loads: #('rich text exporter' 'parser');
   repository: 'github://pillar-markup/pillar:dev-8/src' ].
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,9 @@ Under writing...
 Similar to markdonw headers are composed of `#` space text on one line.
 The headers can be from 1 to 6
 
-```
+~~~Markdown
 ### Header Level3
-```
+~~~
 There is not support for other forms of declaration. 
 
 ### Anchors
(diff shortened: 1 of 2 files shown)


Ducasse:
I do not get why you want to add parser to the baseline. 
We do NOT want the parser the pillar parser.

olekscode:
If I install Microdown, class comments are not rendering.
Screenshot 2020-07-07 at 00 29 42

I described this problem in issue #41 

I received error messages that `PRPillarParser` is missing. And since it is part of the `'parser'` group, I added that group to the baseline. After that everything was fine.

Screenshot 2020-07-07 at 00 35 26

Adding `'parser'` to the baseline seemed to have fixed the problem.
But I don't know if this is the best solution.

olekscode:
If I install Microdown, class comments are not rendering.
Screenshot_2020-07-06_at_17 43 02

I described this problem in issue #41 

I received error messages that `PRPillarParser` is missing. And since it is part of the `'parser'` group, I added that group to the baseline. After that everything was fine.

Screenshot 2020-07-07 at 00 35 26

Adding `'parser'` to the baseline seemed to have fixed the problem.
But I don't know if this is the best solution.

PR #45, opened 2020-07-07
44 add more math examples to the code comment of mic math inline block

olekscode:


Ducasse:
This creates an loop in the baseline. 
To load microdown with pillar we should load PillarMicro from pillar. 
And in the future we may split it differently

olekscode:
The goal of this PR was to add some math examples to `MicMathInlineBlock`
But because of the problem that I have described in my email, it got mixed with other things.

Screenshot 2020-07-07 at 09 51 23

```Markdown
Here are several examples:

#### Example 1
Code: `$V_i = C_0 - C_3$`

$V_i = C_0 - C_3$

#### Example 2
Code: `$a^2 + b^2 = c^2$`

$a^2 + b^2 = c^2$

#### Example 3
Code: `$\\sin(x) = \\sum_{n=0}^{\\infty} \\frac{(-1)^{n}x^{2n+1}}{(2n+1)!}$`

$\\sin(x) = \\sum_{n=0}^{\\infty} \\frac{(-1)^{n}x^{2n+1}}{(2n+1)!}$

#### Example 4
Code: `$f(a) = \\frac{1}{2\\pi i} \\oint_{\\gamma} \\frac{f(z)}{z - a} dz$`

$f(a) = \\frac{1}{2\\pi i} \\oint_{\\gamma} \\frac{f(z)}{z - a} dz$

```

PR #58, opened 2020-07-20
Fix dependencies so microdown-pillar can be loaded without 

noha:
rich text composer. Dependencies are too broad to make microdown useful on its own. Reduced number of dependencies


noha:
@Ducasse Anything preventing this to be merged?

Ducasse:
Vacation :). 
I will integrate it.

Ducasse:
Trying to revert.

PR #103, opened 2020-11-16
Loading script from README doesn't work

jvalteren:
Hi!
The Metacello load script in the README does not work. I see 2 causes:
1. The branch 'master' no longer exists, which causes a 'Revspec master not found' error.
2. The Microdown baseline contains a dependency on baseline PillarDocumentModel using git tag v8.0.12, which doesn't contain this baseline class.

I just noticed there is a BaselineOfMicrodownDev class, which solves problem 2. However, this is not mentioned in the README. 

Nice work by the way!

Kind regards,
Jonathan

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Such packages should be moved in the future to other location (probably pillar i
 ```
 Metacello new
   baseline: 'Microdown';
-  repository: 'github://pillar-markup/Microdown/src';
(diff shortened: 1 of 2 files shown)


Ducasse:
Thanks I did not see your PR before. :(

Ducasse:
This is fixed.

PR #121, opened 2021-05-11
correction skip test

leofrere:


Ducasse:
I did not fully understand but the tests are easy to fix later

PR #123, opened 2021-05-14
add positive and negatif tests 

leofrere:


Ducasse:
Looks good I will do a pass on all the tests and start to plug the inline parser later.

PR #140, opened 2021-05-27
MicPillarConverter

leofrere:
finish to migrate asPillar into a visitor with tests.
I change asPillar for MicRawBlock.


Ducasse:
I will integrate and repackage it because asPillar should not be in microdown package.

PR #142, opened 2021-05-28
avoid an error if the value is not a String

leofrere:


Ducasse:
what would be good is to have a test that covers this case.

PR #144, opened 2021-06-02
move up inspectionTree in MicElement

leofrere:


Ducasse:
The granularity of your PRs is improving :)

PR #145, opened 2021-06-02
extract method fixBlockWitoutMicText

leofrere:
- put tests in differents protocols
- extract method fixBlockWitoutMicText
- add wrappedElement with tests

### DIFF
--- a/src/Microdown/MicInlineElement.class.st
+++ b/src/Microdown/MicInlineElement.class.st
@@ -161,3 +161,8 @@ MicInlineElement >> substring: aSubString [
 MicInlineElement >> text [
  	^ substring
 ]
+
+{ #category : #accessing }
+MicInlineElement >> wrappedElement [
+	^ children first
+]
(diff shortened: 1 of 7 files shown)


src/Microdown/MicInlineElement.class.st
```diff
@@ -161,3 +161,8 @@ MicInlineElement >> substring: aSubString [
 MicInlineElement >> text [
  	^ substring
 ]
+
+{ #category : #accessing }
```
Ducasse, line 165: I'm not sure that this is always true. May be we will add a new superclass to Bold, monospace, Strike and italic and we could move this method there but for now this is ok.

src/Microdown/MicMicrodownTextualBuilder.class.st
```diff
@@ -202,7 +202,7 @@ MicMicrodownTextualBuilder >> figureURL: aUrl withCaption: aDescriptionBlock wit
 		ifNotEmpty: [ 
 			self rawFigureDescription: aDescriptionBlock.
 			self raw: parserClass urlOpenerMarkup.
-			self raw: aUrl asString.
+			self raw: '/', aUrl segments first.
```
Ducasse, line 205: We will revisit this when we will start to load figure from file.

Ducasse:
Nice to have less skip!!!

PR #147, opened 2021-06-02
remove tests on script parameters

leofrere:
remove tests on script parameters because there is not parametersBlock in Microdown


Ducasse:
Indeed in fact I do not really get what are these tests.

PR #151, opened 2021-06-04
parse: can parse text as arument

leofrere:
resolve the probleme when parse: totaly break when a Text is passed as argument


src/Microdown/MicroDownParser.class.st
```diff
@@ -655,7 +655,9 @@ MicroDownParser >> parse: aStreamOrString [
 	current := root := self newRootBlock
 		setParser: self;
 		yourself.
-	inStream := aStreamOrString readStream.
+	inStream := aStreamOrString isStream
```
Ducasse, line 658: We could also do it like that aStreamOrString asString ..... systematically. 
This avoids the tests and the condition. If this is a string then it returns self.

PR #155, opened 2021-06-09
add object to converter

leofrere:


Ducasse:
Done in previous PR :)

PR #158, opened 2021-06-16
put newLine before header, paragraph, figure, list and codeblock

leofrere:
put newLine before header, paragraph, figure, list and codeblock and correct factory method of these elements.


src/Microdown-Tests/MicMicroDownSnippetFactory.class.st
```diff
@@ -129,84 +134,96 @@ This is a codeBlock
...
 
-	^ self headerLevel: 3
+	^ String cr, (self headerLevel: 3)
 ]
 
 { #category : #headers }
 MicMicrodownSnippetFactory >> headerLevel3SampleWithNewLine [
 
-	^ (self headerLevel: 3), self newLine
+	^ String cr,(self headerLevel: 3), self newLine
```
Ducasse, line 208: better use self newLine

PR #160, opened 2021-06-18
Convert pillar book to microdown book

leofrere:
the PR has a link with an orther PR in Pillar with the same name


src/Microdown/MicAbstractMicrodownTextualBuilder.class.st
```diff
@@ -96,6 +98,11 @@ MicAbstractMicrodownTextualBuilder >> raw: aString [
 	stream << aString
 ]
 
+{ #category : #initialization }
+MicAbstractMicrodownTextualBuilder >> setLine: aLine [
```
Ducasse, line 102: I would call it setNewLineCharacterString:

PR #171, opened 2021-07-19
Feature/mic trasnformer

leofrere:
- add transformers for Microdown
     - CodeBlockEvaluator (Microdown - Transformer)
     - MicFileInclusion (Microdown - Transformer)
     - MicDocumentListTransformer (Microdown - HTML)

- add new visitor
    - MicHeadReunbering (Microdown - NewVisitor)
    - MicCodeblockLineNumber (Microdown - NewVisitor)
    - MicHTMLWriter (Microdown - HTML)


Ducasse:
I revert this PR because we have a broken system. 
for example, Transformer should not depend on ChrysalConfiguration (this is a non-sense)

PR #187, opened 2021-07-20
fix tests evaluator

leofrere:


Ducasse:
One point per PR. Because I was touching the visitInputFile too.
So we will have a conflict.

Ducasse:
So since two topics are mixed I cannot integrate one without the other.

guillep:
Hi Stef, I don't understand, this PR does not touch `visitInputFile`. I only see that it modifies `MicCodeBlock`. Maybe you mean something else?

Ducasse:
May be this is other PR.

PR #192, opened 2021-07-23
add tests for MicNodetransformer + remove visitText

leofrere:


Ducasse:
Thanks this is nice. Could you add a test where we change a header into a codeBlock (to have a test with two different kind of nodes?)

PR #219, opened 2021-09-15
217 rich text composer has skipped tests

kasperosterbye:
There is still one skipped test having to do with problems in the image resolver. I left it skipping for now


Ducasse:
Thanks a lot!!!!

PR #224, opened 2021-09-17
46 problem with backslashes in math environments

### LINKED ISSUE #46: Problem with backslashes in math environments
Backslashes `\` are very important in math environments, but Microdown removes them.
So they have to be escaped, which is very annoying and breaks the compatibility of Microdown math with standard LaTeX math.

For example, in LaTeX I would write
```TeX
$ f(a) = \frac{1}{2\pi i} \oint_{\gamma} \frac{f(z)}{z - a} dz $
```
and get this: 
ComplicatedEquation

But in Microdown it doesn't work:
Screenshot 2020-07-07 at 00 29 42

What I have to do then is escape all backslashes. Like this:
```TeX
$ f(a) = \\frac{1}{2\\pi i} \\oint_{\\gamma} \\frac{f(z)}{z - a} dz $
```

Then everything is fine:
Screenshot 2020-07-07 at 00 35 26

But it would be great if backslashes were not captured in math environments.

kasperosterbye:
fixes #46. In general inline use backslash escaping. Now changed so the inline math is passed verbatim.


Ducasse:
I do not really know what to do so it will take me some more time to see.

kasperosterbye:
My analysis of the situation is the following:

The abstract superclass for inline blocks `MicInlineBlock` adds a method `cleanSubstring` which is called as part of instantiation (in the class side methods `from:to:withString:` and others like it).

So inline blocks has two hook methods, `cleanSubstring` and `closeMe`. `cleanSubstring` is called first.

The `cleanSubstring` method is the one that does the handling of `\`. 

Currently monospace format override the method to do nothing (not expand `\`) with the comment:

> does nothing on purpose

**This PR prevents backslash handling in `MathInlineBlock` (and `MathBlock`). the same way**

I notice that `MicRawBlock` also does `\` handling, and properly should not. The class comment of `MicRawBlock` seems to indicate that this block is deprecated somehow.

kasperosterbye:
I will update this PR to include the above mention issue with rawblock, and include some extra tests for math and raw to make sure they do not re-appear.

Ducasse:
Sorry I was unclear. There was a conflict and I did not know how to fix it.

kasperosterbye:
I went ahead and merged it. I will normally not merge my own requests, but since it was able I did.

Ducasse:
Ok did you check the GH (not the release one because there is a test failing that I do not understand).

kasperosterbye:
I merged it here on github, but that might be bad practice - but there was a huge green button "MERGE" - begging to be pressed....

PR #227, opened 2021-09-20
226 remove new from associations as extension of dictionary

kasperosterbye:
As far as I can see, this is the proper way to handle this thing. No extension method, and idiomatically correct usage of existing library.


kasperosterbye:
This state there are conflicts. I might go ahead at some other time.

PR #230, opened 2021-09-21
229 improve readme

### LINKED ISSUE #229: Improve README
- put the loading at the top (like in most projects) so one does not have to scroll down 
- use Smalltalk syntax highlighting

astares:
Fix #229


kasperosterbye:
I agree this change is a good idea.

There is a concept I call "debugging by documentation". I notice this sentence:


> The following code is not able to display it because markdown quote block are strange and interpret nested block. 

What is the use-case for interpreting the contents of a code-block? Did someone confuse it with a quote-block?

PR #231, opened 2021-09-22
218 - arguments splitter has no tests and is non trivial

kasperosterbye:
I believe this version (a complete rewrite) of `MicArgumentsSplitter` is robust, and has a series of test to go with it.


Ducasse:
I'm thinking that splitOnFirst could be added to Pharo with some extra comments and a couple of tests.
What do you think?

kasperosterbye:
I still do not have a good enough feel for "what goes where", and the "bloat vs. lean" balance. If we add a splitOnFirst, one would also think there should be a splitOnLast. It is a simple thing to do. 

I am leaning towards no adding it.

Ducasse:
My rule of thumb is 
- if the method has a nice functionality.
- a nice name
- encapsulate none obvious and dull code
then it is worth.

Ducasse:
And to me it looks like it is because I would hate to code this.

PR #232, opened 2021-09-23
216 package name microdown microdown rich text composer should be microdown rich text composer

### LINKED ISSUE #216: Package name Microdown-MicrodownRichTextComposer should be Microdown-RichTextComposer
To keep it consistent with the other packages.

There is an existing package name `Microdown-RichTextComposer` which only contains some extension methods.

Also - the test class `MicRichTextComposerTest` is in the same package as `MicRichTextComposer` which is against the conventions also.

kasperosterbye:
Fixes #216. I believe this will fix the package name issue. It does for me.


Ducasse:
It was named like that because there was a MD-PillarRichTextComposer.

kasperosterbye:
He he - well at least there was a good reason

Ducasse:
Oops the changes broke the build.

Ducasse:
I wonder if you changed the baseline?

kasperosterbye:
Well spotted Stef - I did not change the baseline.

kasperosterbye:
The baselines have been changed too now.

PR #235, opened 2021-09-24
fixed baselines too

kasperosterbye:


kasperosterbye:
Forgot the baselines of #216 as you suspected. Now fixed.

Issue #233, opened 2021-09-23
Tables do not render in MicRichTextComposer


kasperosterbye:
I believe it is possible to do, though not completely perfect.
Using FTTables, one can get a table morph, which can then be inserted the same way as images and latex figures.

Being able to render tables to some reasonable level will make much documentation nicer.

PR #236, opened 2021-09-24
fixes #234. And adds some tests for remotely located images

### LINKED ISSUE #234: RichText rendering of figures should not include newlining
The current implementation of `MicRichTextComposer>>visitFigure:` inserts newline and tabs before a caption, even when there is no caption. The implementation should not behave if a block is written. 

One fix is to exit after the figure has been inserted if no caption exist.

kasperosterbye:


kasperosterbye:
After this test I am able to do #233 and render tables in RichText

PR #246, opened 2021-10-14
Moved image cache handling to class side.

kasperosterbye:
In order to support stylers to utilize the image cache, it was moved to the class side.


kasperosterbye:
I was making a plantuml extension, which would like to use the same image cache as latex. But stylers live on the class side, so it was necessary to move it to where it belongs.

kasperosterbye:
I added merged in the #264 into this branch to see if it fixed the problems - and it did.

Issue #262, opened 2021-10-28
MicRichTextComposerTest>>testStylerForSmalltalk fails


kasperosterbye:

Ducasse:
We should go over the skipped tests.

kasperosterbye:
Yes, that is issue #115 to do that...

PR #264, opened 2021-10-29
(MERGE FIRST) Refactoring of code stylers

kasperosterbye:
I refactored the stylers mechanism to sit in a class by itself and not cache compiled methods but class name and selector to get around issue with obsolete classes and other spurious aspects. 

This is done in part to try to track down some errors in #262 i suspect is rooted in that problem.


Ducasse:
Thanks Kasper. I'm restarting to work on MD slowly.
I will integrate the changes I made with Lyna first to clean that plate and I will come back here.

kasperosterbye:
Wonderful @Ducasse . I can't figure out what is going on here.

Ducasse:
I had not time to investigate (correcting exam + PhD). 
Now I added https://github.com/pharo-project/pharo/issues/10294

kasperosterbye:
There is a hard to track down bug in this PR. I will close it now and try a different approach.

kasperosterbye:
I am opening this again as I had a new idea (hack) I wanted to try.

kasperosterbye:
OK, this is what I did:
- changed the name of the code styler pragma from `codeblockStylerFor:` to `codeblockStyler:`, and based the logic on the new name. This prevented interference from the erroneous included methods from the main branch.
- Added two dummy tests to `MicRichTextComposerTest` which were otherwise moved to `MicCodeStylerSpecTest`. Adding those dummy methods prevents the erroneous included methods from the main branch to fail.

PR #268, opened 2021-11-04
fixes #267 - simple support for the file uri.

### LINKED ISSUE #267: MicResourceReference>>fromString: should accept both file: and http:
The current implementation assumes anything not a http is a file path string. 
I believe this should be extended to keep default to be file path string, but also allow for the file uri.

kasperosterbye:


kasperosterbye:
The test fails with this **unrelated** problem:
image

The problem is that the text exporter turns a relative url from microdown syntax into a (wrong) file uri.
Wrong because there is no such thing as a relative uri.

kasperosterbye:
I will close this one - there are a few other things which need to be cleaned up in the resolvers.

PR #272, opened 2021-11-05
The CI chain is not working as it should

kasperosterbye:


kasperosterbye:
The test I added (and which fails) is this one:
```smalltalk
testOldMethodsGone
	self assert: (MicRichTextComposer class selectors includes: #expressionStylerFor:) not
		description: 'MicRichTextComposer still contains method #expressionStylerFor:'
```
It fails if `MicRichTextComposer class` implements `expressionStylerFor:`. And it fails as can be seen:

image

This should NOT happen as one can verify in the [source code of `MicRichTextComposer` in the `dev`or `TestingThatExpressionStylerFor-IsGone` branch](https://github.com/kasperosterbye/Microdown/blob/TestingThatExpressionStylerFor-IsGone/src/Microdown-RichTextComposer/MicRichTextComposer.class.st) do not contain this method.

> **My conclusion is that the CI tests the wrong class, or keeps the wrong methods from the release image (where the method do indeed exist)**

HELP NEEDED!!!! @Ducasse (or someone else)

Ducasse:
Thanks I will try to find someone to help us.

kasperosterbye:
I am now a but uncertain if the whole problem is obsolete classes from the MicRichTextComposer class variable Styler. I think the way to go is to make certain that PR #264 does indeed solve the issue.

kasperosterbye:
I close this issue now as I firmly believe PR #264 fixes the problem.

PR #274, opened 2021-11-05
Explicit add and remove of offending methods (DO NOT MERGE)

kasperosterbye:


kasperosterbye:
Nope - not the way ahead.

PR #280, opened 2021-11-07
Fixes #277.

### LINKED ISSUE #277: Allowing pipe symbol '|' in tables
Tables in Microdown should allow a way to escape `|` so one can write a `|` inside a cell.

(From discussion of issue #50)

kasperosterbye:
Allowed pipe `|` to be escaped in tables using `\|`. Also included a test to support it.


Ducasse:
Just a question why not using '\|'? why do we need 16rFFFD asCharacter asString?

kasperosterbye:
Because the next line splits into cells on the non-escaped pipes. So I need to replace it with something else. And this was the only character I was certain would not appear in any input.

PR #287, opened 2021-11-11
Adding test documents for http resolving

kasperosterbye:
To make sure the http resolving works, I just add these documents to the doc folder (which did not exist currently).


kasperosterbye:
I will merge this myself as it is not code based.

PR #291, opened 2021-11-21
Cleanup and beautify

kasperosterbye:
- Added more facade methods on class side of key classes
- tried to make some method names more consistent 
- added an examples smoke test for RichTextComposer
- Beautiful comments settings removed from RichTextComposer


kasperosterbye:
@Ducasse This should be ready to roll.

Ducasse:
I like the clean up and beautify :)

PR #293, opened 2021-11-23
Text styler refactoring

kasperosterbye:
This PR refactors the textStyling out of MicRichTextComposer into a class by itself which allow you to use MicRichTextComposer to change the size of headers and the style of bullets.


kasperosterbye:
If you have the time @Ducasse. 
Btw, I got column and columns to work in RichText as well as self scaling images (in an other branch). It is still a little rough, but I do believe that it will be possible using this PR to make a MicroDown slide presenter that runs in the image. This PR is needed to get bigger headers (but not all the time).

Ducasse:
Nice. I'm travelling tomorrow and teaching thursday then talking to people :).
This is nice that you moved some of the class side method to the instance side.

Issue #298, opened 2021-11-26
Delimiter for paragraph


Ducasse:
Right now a paragraph is ended by a newline and Pavel was  telling me that it would be better to have cr. 

I do not know but I took note about it. 
We can discuss this point.

Ducasse:
In macDown this is the same as ours. 

in GH

this is a new pargarpha wiht super mogjk kjh jkh jkh kjh kjh kjhkjh kjh kjh kjh kjhkjh kjh kjh kjh kjh kjhkj hk kjh
jhhjhjhkjhkjhkjh
jkhkjhjkhkjhjkhjk

khjkjhjkkjhjkjkhj

kasperosterbye:
Most paragraphs are so short I do not know if list or string is better - I am fine with both. You need to concatenate the lines into a list before doing the inline parsing as e.g. bold can stretch over multiple lines
this is **bold
and so is this** is what I mean

I do not understand the comment about newLine as paragraphs are ended by String cr as far as I can see. Is he refering to the Text generated by MicRichComposer? Or is it the latex generator? Or something else?

Ducasse:
Good remark for bold.
Now it means that 
```
this is **bold
and so is this** is what I mean
```
should be represented as 

```
this is **bold and so is this** is what I mean
```
because you want to render paragraphs as a nice block

```
jkhkjh kjh jkh kjh kjh kjh kjh kjh kjh kjh kjh kjh kjh iuy iuy
 uyf y jk hj lkklj ;kl  kl;k  kl; ;lkkjgjhg g f hf
 hfhf hgf hgfhgfhgfhgfhgfhgf
 hgfhg fhgf hgfhgf 
```
should be renderer as 

jkhkjh kjh jkh kjh kjh kjh kjh kjh kjh kjh kjh kjh kjh iuy iuy uyf y jk hj lkklj ;kl  kl;k  kl; ;lkkjgjhg g f hf
 hfhf hgf hgfhgfhgfhgfhgfhgf hgfhg fhgf hgfhgf 

and not four lines stopping before the end. 

So now I understand what we should do.
We keep the delimiter of paragraph as it is: an empty line like GH and others MD.  

Now the paragraph text should not be managed as a concatenation of line + one crs. 
Because this is the job of the renderer to decide and per default the renderer should take the line and produce a paragraph without cr in the middle. 

In addition I would like to have more newlines emitted in the micRichTextComposer to clearly see the paragraph.

[...]

PR #299, opened 2021-11-26
#298 Fixing the cr logic in paragraph

Ducasse:


src/Microdown/MicParagraphBlock.class.st
```diff
@@ -30,7 +30,7 @@ MicParagraphBlock >> addLineAndReturnNextNode: line [
 
 	text := text
 		ifNil: [ line trim ]
-		ifNotNil: [ text , String cr , line trim ].
+		ifNotNil: [ text , String space , line trim ].
```
kasperosterbye, line 33: I am not happy about this thing happening in the parser, as it throws out information. 
In github markdown, new lines in input are respected, and I think one need very good reasons to make it impossible to make a renderer which mirrors github output.

kasperosterbye: Rather I could imagine composers to merge lines if needed. I believe it is the visitText: methods in MicRichTextComposer which need to remove cr's. This could be done as:
```smalltalk
visitText: anInlineBlock
	canvas << (anInlineBlock substring asText copyReplace: String cr with: textStyler intextLinemerger)
```
moving the actual behaviour to the text styler which can be set before the composer starts to compose.

For the latex version I guess something similar can be done.

src/Microdown-Tests/MicEnvironmentBlockTest.class.st
```diff
@@ -248,9 +248,7 @@ the paragraph.
 ?>'.
 	self assert: root children size equals: 1.
 	env := root children first.
-	self assert: env children first text equals: 'Here we are a paragraph and the paragraph should not eat the end of the
-environment, else we force writer to enter a new line to declare the end of
-the paragraph.'.
+	self assert: env children first text equals: 'Here we are a paragraph and the paragraph should not eat the end of the environment, else we force writer to enter a new line to declare the end of the paragraph.'.
```
kasperosterbye, line 251: Ahh, this was most enlightening.

Issue #302, opened 2021-11-27
Better control with line and paragraph breaks.


kasperosterbye:
In both Latex and HTML there are mechanisms in place that allow more fine grained control how spacing between paragraphs is done.
Microdown is intended to be used for both books, slides, in-image documents and online documents. They all have different traditions, and in some places one want a compact presentation, in others a more airy one. 

It might be attractive to recognise that many different purposes exist, and find ways in which the writer can control the resulting layout better. 

Two solutions might be:
- A default compact presentation, with explicit commands for newline and vertical space.
- styleSheets which allow consistent but different layouts to be used

I am sure there are other ways to do it, but these were the two I could think of.

kasperosterbye:
The solution I propose now is to do a default stylesheet with a paragraph breaking layout as close to github as possible.

kasperosterbye:
Merged and closed

PR #307, opened 2021-12-02
Text styler and better linebreaks

kasperosterbye:
This PR makes the interblock spacing as close to github as possible. It does so by the text styler. Tests have been updated to account for this change. 
There is one new skipped test because there was no test for quoteblock, and quoteblock is not implemented in MicRichTextComposer.
All classes in Microdown-RichTextComposer have been moved under a Composer tag. Originally I was working on getting a presenter component done, but got stuck in rubric problems. I expect to be able to finish the Composer part as soon as pharo PR #10536 is fully checked.


kasperosterbye:
I will look at the fail later today.

kasperosterbye:
I give up!

The sources in the class `MicRichTextComposer` in this branch `TextStylerAndBetterLinebreaks` I try to merge, DO NOT contain a method named `testNestedList`. It there sometime earlier, but was obsoleted.

I continue to get errors regarding stuff that was once there, but has long since gone. My only guess is that the build used by the CI is peculiar, and I have no idea of how to solve it.

kasperosterbye:
I used the previous trick and just installed a dummy method with the same name which prevent the old method from existing.

kasperosterbye:
Fixes #302 and fixes #298 - But take a look @Ducasse,I believe it does what your PR proposed, just incorporated into the text styler so one can create a different style - and keep a style used for testing so style changes do not invalidate a lot of the tests.

PR #308, opened 2021-12-02
Fixes #300 offline handling

### LINKED ISSUE #300: Offline handling
I had an email exchange with @Ducasse regarding handling of resources when the image is offline. I move it here as there is an issue to be addressed. The issue arose as I deleted the Offline feature in MicRichComposer when I refactored it.

Stef wrote about the requirements:
Ideally 
- a load image resources when online should show "an image not found on at http://www.pharo.org/…."
	In fact I produced an image morphi with such text
	
- a load image resources when offline should show an image with “you should see  http://www.pharo.org/…. 
   when you will be back to connectivity”

I agree with Stef on this.

Those changes are relatively easy do do, and are local to `MicHTTPResource`. 

An other aspect of the removed Offline aspect is making some images accessible for access when the image is offline. Stef wrote:
> OffLine was important because the comments can look good and we can prepare all the images
because shipping an image and it will still display. 

Again, I agree that this is a good feature to have.

The original solution would not achieve this as the cache was an LRU and not an image repository. In the current solution it is possible to use images (forms and morphs) using the pharo uri. `pharo:///Class/selector/arg1/arg2/arg3`. If one wants a set of preloaded image forms to be accessible from a preloaded repository one could for example write:

```smalltalk
![Cool logo](pharo:///MicImageRepository/getImageForm:/coolLogo width=85%)
```

kasperosterbye:
Fixes issue #300 on the microdown side.

What is needed now on the beautiful comments side is to reintroduce the Offline settings, and the logic discussed in this #300 issue. But Microdown should now be able to accommodate the required behaviour, without it being dependent of Beautiful comments nor including settings which really belong to beautiful comments.


kasperosterbye:
Too many commits were included...

PR #310, opened 2021-12-06
300 offline handling

kasperosterbye:
This PR prepares Microdown for BeautifulComments to do its Offline handling


kasperosterbye:
Not ready yet - working in tandem with a BeautifulComment PR to make sure they play together.

kasperosterbye:
Ready to roll. When you have time @Ducasse. It is not a showstopper for me to get this one merged. It works with Beautiful comments PR #14, and gives Offline settings to Beautiful Comments.

src/Microdown/MicHTTPResource.class.st
```diff
@@ -34,16 +35,24 @@ MicHTTPResource class >> imageCache [
...
+	ImageNotCachedBlock := nil
 ]
 
 { #category : #images }
 MicHTTPResource class >> rawImageFrom: aURL [
 	[ ^ ZnEasy getImageOfType: (ZnMimeType main: 'image' sub: '*') fromUrl: aURL ]
-	on: Error
-	do: [ ^nil ]
+	on: Exception
+	do: [ ^ nil ]
```
Ducasse, line 55: I was thinking that we could just return an image so that the method always returns polymorphic objects.

kasperosterbye: The nil is used to let the renderer rather than the ressource decide what to do when no image is found. I believe this is the best way.

src/Microdown/MicHTTPResource.class.st
```diff
@@ -7,24 +7,25 @@ Class {
...
-	
-	ImageCache at: urlString put: aForm
+	(self imageCache includesKey: aUrl)
+		ifTrue: [ ^ self imageCache at: aUrl ].
+		
+	ImageNotCachedBlock 
+		ifNil: [
+			^ self imageCache at: aUrl put: ((self rawImageFrom: aUrl) ifNil: [ ^nil ])]
+		ifNotNil: [ 
+			^ ImageNotCachedBlock cull: aUrl]
```
Ducasse, line 28: Why do you use cull: and not value: because cull: is costly and reflective so I really tend to use it only when absolutely needed.

src/Microdown/MicHTTPResource.class.st
```diff
@@ -7,24 +7,25 @@ Class {
...
-
-{ #category : #accessing }
-MicHTTPResource class >> imageAt: urlString put: aForm [	
-	"the cache contains results of for example ZnEasy getPng:, i.e., a form"
-	
-	ImageCache at: urlString put: aForm
+	(self imageCache includesKey: aUrl)
+		ifTrue: [ ^ self imageCache at: aUrl ].
+		
+	ImageNotCachedBlock
```
Ducasse, line 24: I found the logic with nil instead of boolean for ImageNotCachedBlock difficult to follow. 
I could not tell you if ifNil is caching or not. 
BTW you do not need the return in the ifNil: and ifNotNil: block

src/Microdown/MicHTTPResource.class.st
kasperosterbye: I believe it was a leftover because I allowed to use a block that could just return a image without using a url - like the box with a "?"

src/Microdown/MicHTTPResource.class.st
kasperosterbye: I tend to agree that is a bit strange - I tried to tell what it was by adding "Block" at the end. It is a way in which HTTP resource avoids supporting Beautiful comments directly, but to give BC the functionality it need to do its offline stuff. The actual block is set in BC as part of its Offline handling, and is nil by default.

kasperosterbye: I will look at the ifNil thingie

Ducasse:
Thanks Kasper!!!!

src/Microdown/MicHTTPResource.class.st
kasperosterbye: I changed it to use value. I had not looked into cull.

src/Microdown/MicHTTPResource.class.st
kasperosterbye: I am not convinced the polymorphism should sit here. I will leave it as is.

kasperosterbye:
OMG - that was a lot of interesting errors in the dev branch. Happy hacking @Ducasse!

kasperosterbye:
@Ducasse @tesonep - I fixed the dependency on PillarCore (and reverted MicRichTextComposer to use the stylers again). 
I did that in PR #319. I merged it myself wanting to see if it would indeed fix the CI on the other failing PR's.
Now the PR's are green.

PR #320, opened 2021-12-20
Refactor resource resolution

Ducasse:
Result of two days of pair programming with guille...
Refactored a lot.


kasperosterbye:
Wow - that is one major job you guys undertook! Kudos!
You want me to take a look at what might be the problem with the obsolete/undefined class?

kasperosterbye:
It was a simple fix for an error I had seen before. 

I think this type of error should be prevented from happening somehow.
One scenario for the error to happen is this:
1. class B has A as superclass
2. a new class A2 is built and tested to substitute for A
3. A is removed
4. A2 is renamed to A

However, in step 3, B superclass become AnObsoleteA, and remains so even after step 4. 

Not sure this is what happened here, but it might have been the case.

kasperosterbye (approved):
I was able to install this PR in a fresh image using the standard Microdown baseline. All tests green. I say merge it.

Ducasse:
Cool. Thanks for the feedback.

PR #321, opened 2021-12-26
Microdown presenter

kasperosterbye:
This implements a simple read-only presenter spec component which is able to render the `Text` produced by MicRichTextComposer.

It is a Spec triplet, Presenter, MorphicAdapter and Morph.

95% of it is about making sure `Text` is rendered correctly and should in my view be done in Spec. But it is not, so now it is here. 

There is no switch back and forth between Text presentation and Microdown editing. Primarily because I can see many situations where that is not needed, but also because it can be done in different ways - switching back and forth, view side by side, or edit microdown using an external editor.

It splits the package `Microdown-RichTextComposer` into two tags, `Composer` and `Presenter`. The same is done in the test side.


kasperosterbye:
Merry Christmas and happy hollidays @Ducasse. I am in no hurry on this one. I will need to spend some time on my sheep program and some other in-house programming.

Ducasse:
Ok I will merge and I will package the Presenter part in a separate package. 
Thanks

PR #322, opened 2022-01-06
Fixes some category naming issues

kasperosterbye:
Someone had left a category name 'utilscaca' in the code. Renamed. Also the main Pharo have been pushing for consistent spelling of 'utilities' rather than 'util' or 'utils'. Fixed those also for combined names like 'private utilities'.


Ducasse:
It was me because such rules pissed me off a lot. 
And I hate losing time with such points. utils vs utilities for example is just insane.

PR #330, opened 2022-01-08
329 environment parses wrong

kasperosterbye:
It turned out that Paragraphs had an error in its accept next line. Simple fix.


kasperosterbye:
Please check at your convenience @Ducasse, it is not a show stopper.

Issue #292, opened 2021-11-21
RichText image links do not work


kasperosterbye:
The generated (rich)text for `[link to pharo](https://pharo.org)` renders as a text with a link that opens the Webbrowser, whereas an anchor with an embedded image do not work: 
`[![](https://pharo.org/web/files/pharo.png)](http://pharo.org/)`, 
though the image do render correctly.

I have a bad feeling that this might be a rubrics error as the attributes in the runs of the text seems correct.

kasperosterbye:
It turns out there are two error in rubrics:
- If an image is the last element in a TextAction, the image do not react to click. 
- Adding a space after the image (so it is no longer the last element) causes an index out of range in rubrics.

Adding a zero width no-break space (unicode FEFF) after all images "solves" the problem.

PR #331, opened 2022-01-09
Hack that fixes issue #292

kasperosterbye:
See comment in issue #292 about what is the problem and its hacky solution.


kasperosterbye:
Please check at your convenience @Ducasse, it is not a show stopper.

PR #334, opened 2022-01-09
redone RichTextComposer examples.

### LINKED ISSUE #332: Examples in RichTextComposer are broken
There is a problem with SpFont but I do not really get what is the problem.

kasperosterbye:
Fixes #332


kasperosterbye:
I reduced the number of examples, and made the text in the examples explain what is going on

PR #335, opened 2022-01-12
333 set the setting to use the resource cache break some tests

kasperosterbye:
The division of labor between BeautifulComments (BC) and Microdown (MD) is (in my understanding the following).

## Microdown
Is responsible for (in its collaboration with BC) to turning a microdown markuped string into a Text object which can be used in Calypso.

## BeautifulComments
Is responsible for composing a microdown string for MD to turn into a Text, and for handling the necessary hacks into Calypso to do this.

## Settings
I have been able to detect the following aspects which should be controlled by settings
1. Rendering. If false - Comments should not be rendered as Text, but as raw strings in Calypso. Default is true.

2. CaptureErrors. If there is an error in the rendering of a microdown string, BC will capture it, and render the string rather than the wrong Text. If this setting is false, a debugger will open. Default is false. There is a smoke test in `MicParsesAndRendersAllCommentsTest>>MicParsesAndRendersAllCommentsTest` to check if any current comments cause problems. It is green as of now.

3. IsCachingRessources. Indicates if the cache in MicAbstractResolver should cache ressources or not.

4. Offline. If true, the only ressources which can be used are those from the cache. If false, ressouces will be loaded into the image from their external location (and stored in the ressource cache if IsCachingRessources is true).

### Location of Settings
Settings Rendering and CaptureErrors are stored as class variables in BeautifulComments class. The settings IsCachingRessources and Offline makes sense also outside the use of Beautiful comments and belong to Microdown. 

The microdown class actually using IsCachingRessources and Offline is MicAbstractResolver. However, that is a rather deep implementation detail, and I moved the actual settings to sit in the class Microdown. AbstractResolver is then fetching the settings from there.

### Settings in the settings browser
The system settings methods are defined in the two classes BeautifulComments and Microdown. They are grouped together in a group made by Microdown and used by BeautifulComments.


kasperosterbye:
Tested and ready to roll, but read the large comment above and review the changes @Ducasse.

Ducasse:
Sounds good!

PR #337, opened 2022-01-14
moved the presenter to its own package Microdown-RichTextPresenter

kasperosterbye:
I have not updated the baseline to include it. That is on purpose as it is still a little unclear where is should end up in the grander scheme of things.


kasperosterbye:
@Ducasse - it seems like a massive change to what is being tested has taken place. A lot of failing test. Should I be worried? Should I help?

kasperosterbye:
Massive amounts of tests broken - investigating.

kasperosterbye:
Seems ready to roll @Ducasse. The huge number of errors was because the presenter was no longer loaded, but its tests were. Now the Presenter-Tests are in its own package `Microdown-RichTextPresenter-Tests`, and not part of the baseline.

Ducasse:
Ok thanks I will redo a pass when I will integrate it because I liwk to have the tests loaded.

PR #357, opened 2022-02-12
Fixes the latex problem due to codecogs service changed their syntax

kasperosterbye:
I have green now - lets see if it works here too


kasperosterbye:
Check when you have time @Ducasse

PR #370, opened 2022-02-24
Html tests

hernanmd:
Add #argumentAt: and #argumentAt:ifPresent: to easily access figure properties through  HTML #addArguments:
Add visitor methods for figures and fix HTML generation according to W3 spec.
Remove skip in some HTML exporter tests.
Add #figureSampleWithoutSizeAndLabel.
Add missing named HTML entities from W3C
Fix MicFigureBlock #printOn: failing when url is nil.
Add MicHTMLTag argumentAt: methods


src/BaselineOfMicrodown/BaselineOfMicrodown.class.st
```diff
@@ -44,7 +50,7 @@ BaselineOfMicrodown >> baseline: spec [
 				with: [ spec requires: #( #'Microdown-Evaluator') ];
 			
 			package: #'Microdown-HTMLExporter'
-				with: [ spec requires: #( #Microdown #'Microdown-LaTeXExporter') ];
+				with: [ spec requires: #( #Microdown #'Microdown-LaTeXExporter' 'NeoCSV') ];
```
Ducasse, line 53: Why a dependency to neocsv?

hernanmd:
It's in the commit comments: Add NeoCSV to read W3C named entities CSV

Hernán

src/BaselineOfMicrodown/BaselineOfMicrodown.class.st
hernanmd: From [A tutorial on character code issues](https://web.archive.org/web/20120513165012/http://www.cs.tut.fi/~jkorpela/chars.html): In the [HTML language](https://web.archive.org/web/20120513165012/http://www.cs.tut.fi/~jkorpela/html-primer.html) one can use the notation &Auml; for [the character Ä](https://web.archive.org/web/20120513165012/http://www.cs.tut.fi/~jkorpela/latin1/3.html#C4). In the official HTML terminology, such notations are called [entity references (denoting characters)](https://web.archive.org/web/20120513165012/http://www.htmlhelp.org/reference/html40/entities/). It depends on HTML version which entities are defined, and it depends on a browser [which entities are actually supported](https://web.archive.org/web/20120513165012/http://www.htmlhelp.com/faq/html/all.html#entity-or-number).

Generated HTML contains special characters with specific format (https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Character_entity_references_in_HTML)[https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Character_entity_references_in_HTML].

The included CSV contain the mappings to convert characters to the entity references.

kasperosterbye:
The issue with the html character issues is actually well spottet!

That does not mean I fully understand we have a problem though. 

We are generating HTML, not parsing it. In that case, I do not understand the need to convert manifest characters like `⎶` to `&bbrktbrk;`. If someone wrote/generated `&bbrktbrk;` directly in the microdown string the html generator can ignore it also because the browser which will read it in the end know how to render those `&...;` names.

If we were trying to convert HTML strings to microdown we do indeed have an issue.

The only two characters which have an issue is (I think) `<` and `&`. But those are handled by `ZnHtmlOutputStream` which I presume you are using in the first place?

hernanmd:
I checked some docs and [posts](https://stackoverflow.com/questions/9587056/should-all-accented-characters-use-html-entities) about this.

It seems that as long as no-one is interested to edit the generated HTML and we can ensure the document has a meta tag specifying UTF-8, then we can omit the character to HTML entity conversion.

I will remove the conversion and add a meta tag with UTF-8 if there is no objection.

kasperosterbye:
@hernanmd 
> I will remove the conversion and add a meta tag with UTF-8 if there is no objection.

Awesome!

hernanmd:
Done. Changes already PR.

PR #373, opened 2022-02-24
Fixes #341 inline code monospace does slash treatment wrong

### LINKED ISSUE #341: Inline code (monospace) does slash treatment
Sample is `Foo\Bar` gets converted to `FooBar`.

Error should be fixed and documented to be fixed by a test in MicInlineParserTest

kasperosterbye:
One change and three tests added


Ducasse:
indeed!!!

PR #374, opened 2022-03-27
DO NOT MERGE!

kasperosterbye:
DO NOT MERGE!
This is just to get a CI feedback.


kasperosterbye:
I have done a major rewrite of the resolver part of Microdown.

The key changes are:

- All resolution is done using Zinc, file as well as http.
- I removed the global variables I originally used to adress the difference between disk and memory filesystem. 
- Disk file system is default, memory file system can be passed explicitly.
- File and Http access to resources might fail for a number of reasons. Failures are all translated into a MicResourceReferenceError, which can catched to do different things. 
- I have made a number of tests to try to ensure it all works, but also deleted a number of tests for classes that has been removed.
- RichTextComposer, Latex, and HTML tests are all green now.
- A significan number of classes in the Core has been obsoleted and removed.
- I have updated class comments to try to explain the inner working of ResourceReferences and the resolver, as well as the input file importer.

kasperosterbye:
I have added `MicDocumentBrowser` back into Microdown. It is part of this PR, and in the baseline. It does not quite work yet as I am adjusting it to use the `MicResourceReferences`.

kasperosterbye:
I believe this PR is ready to roll. When you have the time @Ducasse

kasperosterbye:
My gosh, now it fails because my tests accesses the github api more than some rate limit allows. 
It works on the linux part but not at the mac, because the linux part ate most, but not all of the rate.
I need to get in touch with the iceberg developers to ask them how they avoid that issue.

src/Microdown-DocumentBrowser/MicPharoPackageCommentResourceReference.class.st
```diff
@@ -0,0 +1,208 @@
...
+	#superclass : #MicPharoCommentResourceReference,
+	#instVars : [
+		'kind',
+		'categoryName'
+	],
+	#category : #'Microdown-DocumentBrowser-ResourceModel'
+}
+
+{ #category : #private }
+MicPharoPackageCommentResourceReference >> browserIcon [
```
Ducasse, line 37: Hi kasper 

I would package browser Icon in the doc browser package to avoid ui in the core. 

S

src/Microdown-DocumentBrowser/MicPharoPackageCommentResourceReference.class.st
```diff
@@ -0,0 +1,208 @@
...
+{ #category : #accessing }
+MicPharoPackageCommentResourceReference >> categoryName [
+
+	^ categoryName
+]
+
+{ #category : #private }
+MicPharoPackageCommentResourceReference >> childrenOfAllCategories [
+	"I return the children of 'comment://package/' "
+	^ (SystemOrganizer default categories
```
Ducasse, line 59: pay attention that this categories stuff should disappear in the future. 
It will take some years so we have time.

kasperosterbye: OK, I will wait :-)
The reason I put it in is simply to get some deeper level of nesting, not the least to support that all microdown projects are grouped together.

src/Microdown-DocumentBrowser/MicPharoPackageCommentResourceReference.class.st
kasperosterbye: The class `MicPharoPackageCommentResourceReference` is actually a model class in the browser package, not the core. The browser package adds `browserIcon` method to some of the core classes, but as extension methods.

PR #377, opened 2022-04-12
Fixes #376 speed is becoming an issue

### LINKED ISSUE #376: Speed is becoming an issue
On the doc folder of Pharo there is a file named `gen_refactoringframework.md` which is nearly 3000 lines long.
Parsing it on my mac takes around 0.35 sec, and richtextcomposing it takes around double - 0.7 sec. Both together around 1 sec. 
It is sufficient to notice a pause of one browse through some files.

So far I believe only common sense has been used to make parsing and rendering run. At some point it might be useful to see if there is a silly bottleneck somewhere.

kasperosterbye:
## Parsing
Using the contents of `gen_refactoringframework.md`, and the call `Microdown parse: contents`.
Timing using:
```smalltalk
((1 to: 25) collect: [:dummy | Time microsecondsToRun: [Microdown parse: contents]]) min
```

Initial result: 305,196 µs
read: non-recursive: 292,053 µs (non-significant - kudos Pharo!)
making identifyMarkupFor: check only in relevant cases : 53,211 µs
Speedup: 5+ times.

Other speedup might be possible, but using `TimeProfiler` there are no 
obviously bad code to attack (few percent gain).

## Rendering
Sadly the `TimeeProfiler` points to `SHTextStyler`, the code styler for pharo.
Rendering all code as raw text gives a speedup on rendering time of 8+ times.

### DIFF
--- a/src/Microdown-RichTextComposer/MicTextStyler.class.st
+++ b/src/Microdown-RichTextComposer/MicTextStyler.class.st
@@ -59,7 +59,7 @@ MicTextStyler >> headerFontSizes [
 MicTextStyler >> headerLevelFont: level [
 	"I return a font annotation to be used with Attributes"
 	headerFonts ifNil: [ self computeHeaderFonts  ].
-	^ headerFonts at: level
+	^ headerFonts at: (level min: 6)
 ]
 
 { #category : #'composer styles' }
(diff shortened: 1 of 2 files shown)


kasperosterbye:
While I have not been able to make the text composer faster, I believe the gain in the parser is worth it.

Ducasse:
Definitively!!!

PR #381, opened 2022-04-22
Beamer writer + test + fixing snippet marckup in example

DelGaylord:


Ducasse:
merge beamer and next point is to update the baseline to take care of them.

PR #386, opened 2022-04-26
Fixes #207, and includes test not to introduce elsewhere

### LINKED ISSUE #207: Rename MicroDownParser as MicrodownParser and same for others
This is currently impossible to save the package once we rename the class.

I will need help from some git expert. Because on mac there is no difference between Microdown and MicroDown so git does not see the difference.

I should probably refactor the code. 
Then force git from the command line to push the changes.

S

kasperosterbye:


Ducasse:
Oh this is great that you succeeded to do it because I could not on my machine and I was super annoyed by the D.

kasperosterbye:
The clue is to rename the file using `git mv MicroDownParser.something MicrodownParser.something` in the terminal

PR #387, opened 2022-04-27
Fixes #384 improve microdown api for doing resolution

### LINKED ISSUE #384: Improve Microdown API for doing resolution
Extract from email between Stef and Kasper

### Stef:
I was wondering if we would not introduce an API at the level of microdown to avoid to have 
direct reference to MicZincPathResolver.
So we could have 

	Microdown >> resolveString: aString withBase: aBase
		
		| doc| 
		doc := self parse: aString. 
              MicZincPathResolver new 
			 absoluteReference: aBase  resolveDocument: doc. 
		 ^ doc

Let me know what you think.

### Kasper
How about a method in MicElement resolveFrom: aBase, changing the above method to:

	Microdown >> resolveString: aString withBase: aBase
		
		| doc| 
		doc := self parse: aString.
		 ^ doc resolveFrom: aBase

kasperosterbye:
Thanks for the suggestion @Ducasse

### DIFF
--- a/src/Microdown-Tests/MicZincPathResolverTest.class.st
+++ b/src/Microdown-Tests/MicZincPathResolverTest.class.st
@@ -18,9 +18,7 @@ MicZincPathResolverTest >> testVisitFileImage [
 		filesystem: #Bogus;
 		yourself.
 	
-	MicZincPathResolver new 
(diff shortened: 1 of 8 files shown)


Ducasse:
Yes this is much nicer. Thanks

PR #392, opened 2022-04-28
Html generation

hernanmd:
Work in progress for Microdown to HTML generation, tests and partially working examples with CSS.


src/Microdown-HTMLExporter/MicListBlock.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #MicListBlock }
+
+{ #category : #'*Microdown-HTMLExporter' }
+MicListBlock >> listItemText [
```
Ducasse, line 4: Normally you get access to the children.
So no need to access the text

Ducasse:
Hi hernan

- why is a HTMLDoc?
- for the arguments there is a class (with a strange name) that factor all the management of arguments (and this way we have the logic for argument handling in a single place). 
@kasperosterbye knows it name and it would be good (if fit) to use it in the block with url.

hernanmd:
Hi Stef

- why is a HTMLDoc?

I included a class comment to explain the rationale.

- for the arguments there is a class (with a strange name) that factor all the management of arguments (and this way we have the logic for argument handling in a single place). 
@kasperosterbye knows it name and it would be good (if fit) to use it in the block with url.

Ok, I needed a mapping, with some defaults already setup, between Microdown attributes and HTML attributes so that when I receive a "label" in Microdown for example, I write a "class" in HTML as attribute, etc. If that's provided somewhere I'd glad to know.

src/Microdown-RichTextComposer/MicroDownParser.extension.st
```diff
@@ -1,6 +1,6 @@
 Extension { #name : #MicrodownParser }
 
 { #category : #'*Microdown-RichTextComposer' }
-MicrodownParser class >> convertToRichText: aString [
+MicrodownParser classSide >> convertToRichText: aString [
```
kasperosterbye, line 4: What is class side?

kasperosterbye:
Regarding the HTMLDoc, there is no existing mapping doing what you need (if we are talking about `htmlArgumentsMap`).

PR #393, opened 2022-04-28
reworked the Github api to improve caching and rate management

kasperosterbye:
Getting back to green - github api rate management


kasperosterbye:
I will merge this one as it was problems only in the document browser module.

PR #400, opened 2022-05-04
Adding methods to build column and columns + adding a new line to the method for slide

DelGaylord:


Ducasse:
would be good to add some tests to cover the columns and column in textualBuilder.

Issue #404, opened 2022-05-06
RichTextComposer adds extra new lines when closing a indented bullet list


kasperosterbye:
This is formatted wrongly:
```markdown
* Item 1
  * by prefixing the `-` or `*` by two spaces, one indent the list by one level
    * Item 1.1.1
  * Item 1.2
* Againg, the indentation is two spaces!
    * If you indent by say 4 spaces instead of two, it is just seen as a new line in the preceding bullet point
```

kasperosterbye:
Obsoleted

PR #413, opened 2022-05-11
Fixes 404: No empty line emitted for end of nested lists

kasperosterbye:
No empty lines added after sub-lists. I could not see this being visually pleasing in any context.


kasperosterbye:
Fixes #404

PR #420, opened 2022-05-12
Differents change done In MicLaTeXWriter and MicSBALLatexWriter :

DelGaylord:
-  change on figure path from reference path to url in the methods includeGraphicsFor and writeLabelOfExternalLink
  -  adding folderName, templateForConfiguration and writerName for build book


DelGaylord:
I will check why it's doesn't pass the test

PR #421, opened 2022-05-12
Corrected a mistake i made so now the tests are green

DelGaylord:


src/Microdown-LaTeXExporter/MicLaTeXWriter.class.st
```diff
@@ -83,7 +83,7 @@ MicLaTeXWriter >> includeGraphicsFor: aFigure [
 	canvas command
 		name: 'includegraphics';
 		optParameter: [ self writeGraphicsParametersOfExternalLink: aFigure ];
-		parameter: [ canvas raw: aFigure url ]
+		parameter: [ canvas raw: aFigure reference path ]
 	"we should see when the reference is resolved if we really emit well figures in the case of file://figures/logo.png"
```
kasperosterbye, line 87: I am a bit worried about the comment. There is no such thing as relative file reference in Microdown. A relative reference is of the form 'path/to/logo.png', relative references work for all kind of reference types (in particular files and http). 'file://figures/logo.png' is not a relative reference. All references starting with a scheme name are absolute.

PR #450, opened 2022-05-23
Fixes #448 raw does not eat it raw

### LINKED ISSUE #448: raw does not eat it raw
`Microdown parse:  '_aaa{{1_23}}bbb_'` is read as
image

kasperosterbye:
Fixes #448 
Fixed problems with raw not doing raw consumption of input.
Changes the parsing of raw, math ($...$), and anchor reference (*{...}*) so the internal structure is ignored and cannot be seen as closing outer inline blocks.


Ducasse:
Thanks a lot!!!!

PR #451, opened 2022-05-25
add hooks for extensibility and add "Microdown-Macrodown" 

estebanlm:
after talking with Stef, he agreed this is the best way of proceed. 
this PR adds a new parser named `MicrodownStrictParser` (inherited from `MicrodownParser`) and  is intended to implement markdown as close to the spec from here: https://www.markdownguide.org/basic-syntax as is possible. 

Right now, two things where added: 
1. paragraphs that start with an HTML tag will be parsed as "raw paragraphs" and it will contain a raw block.
2. paragraph that contains two spaces or more and an enter will insert a line break block (which in time has to become a <br/> in HTML.


kasperosterbye:
I believe supporting the html tags is an awesome idea. In particular supporting them at both paragraph and inline.
Should we rename this new version to Macrodown - because it is now larger than most marcdown :-)

I have a few minor comments which I will put in the code.

src/Microdown-MarkdownStrict/MdsRawParagraphBlock.class.st
```diff
@@ -0,0 +1,51 @@
...
+ 	^ aVisitor visitRawParagraph: self
+]
+
+{ #category : #'parse support' }
+MdsRawParagraphBlock >> closeMe [
+	| nodes |
+
+	nodes := self inlineParse: text.
+	self children: nodes
+]
```
kasperosterbye, line 45: The notion of raw is not defined, but I view raw as something where the inner part is **not** parsed, or at least not parsed by the microdown parser, but is parsed by something else (latex, verbatim copy, whatever).

kasperosterbye: Perhaps the classname should be MdsHTMLParagraph?

kasperosterbye:
I will not be able to look at this until Monday. You do not deserve this number of failed tests. There are a number of serious invariants which might or might not be broken as noone decided to document the inline parser. In particular, it is assumed that all delimiters are cashed in some shared variables.

PR #466, opened 2022-06-02
Fixes 464 unfinished inline code/monospace block breaks inline parser

kasperosterbye:
Index out of range error fixed


Ducasse:
Oh nice. 
This is a cool that you fix just parts!
Thanks a lot.

PR #468, opened 2022-06-07
Document browser merge to dev

kasperosterbye:


kasperosterbye:
I believe the fail is related to https://github.com/pharo-vcs/iceberg/issues/1588 (change in github rates system).

PR #481, opened 2022-06-09
Fixes 474 multiline mathblock should handle latex comment

### LINKED ISSUE #474: Multiline mathblock should handle latex comment
In order to add an argument list to multiline math without getting in conflict with github we can use the syntax
```
$$%label=formula:fancy&caption=Cool formula
...
$$
```
But for this to work, Microdown must be able to extract argument list from both `$$....` and `$$%....`. 
At the moment it does not.

kasperosterbye:
Fixes #474

### DIFF
--- a/src/Microdown-Tests/MicMathBlockTest.class.st
+++ b/src/Microdown-Tests/MicMathBlockTest.class.st
@@ -61,6 +61,22 @@ V_i = C_0 - C_3
 ]
 
 { #category : #tests }
+MicMathBlockTest >> testMathBlockWithParametersAndLatexComment [
+
+	| root math |
+	root := self parserClass new parse: '
+
+',MathOpeningBlockMarkup,'%key=cite&label=42
+V_i = C_0 - C_3 
+', MathClosingBlockMarkup.
+	math := root children first.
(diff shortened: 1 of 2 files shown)


Ducasse:
Nice :) BTW I will merge it and I would like to have only anchor so like that we have anchor for figure, section and math.
Is it ok for you?

Ducasse:
For a section we put 

```
# my Foo Section
@anchor foo


and later we can have a reference to as in See Section @foo@
```

So I would like to have 

```
$$%anchor=myFooMathEquation
1 + 2
$$
```

instead of 

```
$$%label=myFooMathEquation
1 + 2
$$
```

I should check that reference are actually working.

Ducasse:
```
See *@myFooMathEquation@*
```

kasperosterbye:
Ahh - that kind of anchor. 
First I am not doing anything with them in RichTextComposer (at least not at the moment). 
It seems mostly to be a matter of consistent naming of arguments, which I am only happy to comply with should need arise.

So - go ahead :-)

Ducasse:
I will when I'm coming back to life

PR #482, opened 2022-06-09
Fixes #475 & #476, sync scroll from source to render in document browser

### LINKED ISSUE #475: Document browser should not scroll source when saving

kasperosterbye:


kasperosterbye:
I just merge as only file changed is in document browser.
Fixes #475 & #476

PR #483, opened 2022-06-13
Document browser development

kasperosterbye:


kasperosterbye:
Update of the browser, and only the browser

PR #491, opened 2022-06-15
Creation of MicInputFileResolver + test

DelGaylord:


kasperosterbye:
The implementation change in MicInputFileResolver is correct and the tests tests what they should.

I might rewrite the whole thing at some later point, as the the whole thing is not using the functionality of the ResourceReferences, which could simplify the code. In particular there is no "resolution" as the input file references should already be absolute before doing the input.

PR #497, opened 2022-06-26
Fixes #411 tailoring richtext composer using microdown

### LINKED ISSUE #411: Tailoring richtext composer using Microdown
It should be possible to make a special annotation `format`, which can make document specific text styling. I think something along the lines of:
```text
{!format|
hearder1Size=48&
header1Font=Papyrus&
bullets=-÷·
!}
```
A subclass of `MicTextStyler` which stores this info is needed, which can draw its defaults from `MicTextStyler`.

The parser and extension mechanism works for this syntax, but a new extension class of `MicAnnotationBlock` is needed (`MicRichTextFormatSpecification`?)

kasperosterbye:
This PR is cool, it allow us to adjust font and spacings of the RichTextComposer from within Microdown using an extension.
To get really large header level 1, and larger body font you can write:
```
{!richtext|headerFont1=Lucida Grande;72&bodyFont=Lucida Grande;24!}
# Huge header
and large body
```
image

I think it will be rather useful for slides

### DIFF
--- a/src/Microdown-RichTextComposer-Tests/MicCodeStylerSpecTest.class.st
+++ b/src/Microdown-RichTextComposer-Tests/MicCodeStylerSpecTest.class.st
@@ -18,8 +18,7 @@ MicCodeStylerSpecTest >> setUp [
 MicCodeStylerSpecTest >> testStylerForMethod [
 	
 	| code richText blueLocation textColor |
-	code := '
-codeStylerFor: aString
+	code := 'codeStylerFor: aString
 <codeblockStylerFor: #smalltalk>
 
 ^bobby new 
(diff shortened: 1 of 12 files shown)


kasperosterbye:
Interesting - I assume the ubuntu test fails because the font does not exist. Not good with such a dependency.

kasperosterbye:
Notice - this is done in the richtext composer package, which eventually ends up in Pharo.
It is possible to move it to a separate package if that is a better design.
@Ducasse

Ducasse:
I like the fact that we can have nice font. Now I would like to have them not on the node but a kind of style at the level of the document

kasperosterbye:
It works for the rest of the document, so placing it at top affects the entire document.

PR #501, opened 2022-07-05
Correction of the method extensionClass because it was taking only th…

DelGaylord:
…e direct subclasses and not every subclasses


kasperosterbye:
Very well spottet!
I believe there are extendable blocks which do not use this method. Or did check this?

DelGaylord:
So i was one commit behind on my own repository so i wanted to merge so i would just have this one commit in more but i don't know why i got 6 tests from MicScriptBlocktest which turn red like on the screenshot below:
image
I will check to see why that

Ducasse:
Ah we fixed it both. I will integrate and add some tests for the nesting of environment so that after I fix the implementation.

PR #506, opened 2022-07-12
Support for center and right in rich text

kasperosterbye:
This PR adds two environment extensions which can be understood by the rich text composer, center and right.
It bridges from these to the standard smalltalk `TextAllignment` centered and rightFlush to do the allignment.

As the standard spec Text inspection cannot properly inspect Text properly, I added an inspector to the Text class which uses `MicTextPresenter`.

The inspection extension is placed in the package `Microdown-RichTextPresenter` so it will not end up in Pharo.


kasperosterbye:
It seems I branched of the document browser. It should not affect anything doing this.

PR #508, opened 2022-07-12
Fixes #507 RichTextComposer should not fail on unknown environment

### LINKED ISSUE #507: RichText composer raise error when environment type is unknown
This expression 
```smalltalk
Microdown asRichText:  '
<!card 
stuff
!>'
```
fails.

kasperosterbye:
old error has gotten a test, and been fixed


Ducasse:
we could also fallback on environment when the tag is not found. 
I do not really see the pros and cons of each approach.

kasperosterbye:
The fix was indeed in the implementation of environment. The problem was an accessor name had changed and no test discovered it.

PR #512, opened 2022-07-14
Writing documentation

kasperosterbye:
I finished the last of the Microdown documents I had planned. See `doc/extension.md`
In addition, I added support for two things to have some examples to talk from:
- Footnotes (an extension of inline annotations) are now supported by the richtext composer
- Source code import is now supported as an environment extension. `<!source|class=MicSourceMethodBlock&method=closeMe!>` will produce a codeblock of the specified method. It works nice for writing about code, but is used to explain how extension is done.


kasperosterbye:
One should get this documentation directly by `MicDocumentBrowser open`, it should be available at the end of the document list.

kasperosterbye:
As this PR includes changes to the core, I will ask @Ducasse to look at it when he finds time for it.

PR #517, opened 2022-07-27
Writing documentation

kasperosterbye:
Added documentation for various aspects.
In particular there is now documentation in the doc folder for extension mechanisms.

A few other things snug in too.
- MicRichTextComposer now supports the footnote extension
- A new extension which includes source-code into the microdown document
- A DNU method is added to the general visitor so it will not crash if case of unknown visit extensions.


kasperosterbye:
Some of the changes affect the core classes, so take a look @Ducasse

Ducasse:
Why do we need to catch does not understand?
I really tend to avoid this.

kasperosterbye:
@Ducasse - the scenario with the DNU arrives as such. Someone writes an extension block type named FooBlaBla. The accept: method in FooBlaBla is: 'visitor visitFooBlaBla: self'. However, dumb person (could be me) forgets to make an extension method in microdown visitor, and when you then run an existing visitor on a document using the extension, the existing visitor will break.
So, the proposed DNU makes code robust to such coding errors.
Of cause I was bitten by this - should we rather just accept the DNU error in those cases and force the coding of the extension method.

Ducasse:
Yes I would prefer that it breaks and then I have to fix it. Instead of doing something magic behind the back

kasperosterbye:
I removed the DNU method in Microdown Visitor. 
There is a problem with CI which is related to a api rate system for Github. 
Issues: https://github.com/pharo-project/pharo/issues/11222, https://github.com/pharo-project/pharo/issues/11481, https://github.com/pharo-vcs/iceberg/issues/1588

However, for our purpose I believe this PR can be accepted as it passed on ubunto before the api bug appeared.

Ducasse:
Nice I really prefer to solve such kind of problems by forcing people to write code.

PR #520, opened 2022-07-28
Fixes #519 Microdown directory resolution

### LINKED ISSUE #519: Resolution is wrong in case of filereference to a directory
At the heart of the resolver is `ZnUrl>>#withRelativeReference:`. On my mac there is a directory named `/Users/kasper/tmp`, and the following code:
```Smalltalk
'/Users/kasper/tmp' asFileReference asMicResourceReference uri withRelativeReference: 'foo/my.md'
```
returns `file:///Users/kasper/foo/my.md` but should return `file:///Users/kasper/tmp/foo/my.md`.

Notice, if `/Users/kasper/tmp` is a file, the result is as expected.

Note, if `/Users/kasper/tmp` does not exist in the filesystem, Pharo assumes it to be a file, not a directory.

kasperosterbye:
@Ducasse - I think this PR fixes the issue we found yeasterday

'/Users/kasper/tmp' asFileReference asMicResourceReference uri withRelativeReference: 'foo/my.md' now returns 'file:///Users/kasper/tmp/foo/my.md'.
Fix is done in `MicFileResourceReference>>#fromFileRef:`

### DIFF
--- a/src/Microdown-Tests/MicFileResourceReferenceTest.class.st
+++ b/src/Microdown-Tests/MicFileResourceReferenceTest.class.st
@@ -126,6 +126,8 @@ MicFileResourceReferenceTest >> setUp [
 	file := filesystem workingDirectory / 'toplevel.png'.
 	PNGReadWriter putForm: image onStream: file binaryWriteStream.
 	
+	(filesystem workingDirectory / 'subdir') ensureCreateDirectory.
+	
 ]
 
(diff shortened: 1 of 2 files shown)


kasperosterbye:
CI is still broken - though all tests green locally

Ducasse:
Thanks a lot I will try it in my branch because I should be near a working solution.

Issue #521, opened 2022-07-29
Logic of <! khjkhjjk !> is not robust enough


Ducasse:
We should make sure that it works for

```
<!inputFile|path=blblbl !>

<!inputFile|path=blblbl!>space
```

kasperosterbye:
I have been thinking about making some changes to MicArgumentList:
* The first element (up to | ) must be a single word
* Allow the | to be optional if there is no further parameters, and moving the rest of the line into body
* Trimming the value (between = and & or end)

To accommodate different usage scenarios, I might implement messages to "normalise", especially item 2 and 3, so it is up to the specific block to decide on the details.

kasperosterbye:
I had not realised, but this is actually two issues in one.
1. `<!env...!> bla` breaking
2. `<!inputFile|path=pathToMyDocument.md   !>` should be trimmed.

PR #522 seems to address 1. And 2. for the last argument.

I recommend accepting #522 as it is an improvement overall.

kasperosterbye:
It has been fixed

PR #522, opened 2022-07-30
Make <!environment|path=foobar  !> junk extract correctly environment…

Ducasse:
…|path=foobar

### DIFF
--- a/src/Microdown-Tests/MicEnvironmentBlockTest.class.st
+++ b/src/Microdown-Tests/MicEnvironmentBlockTest.class.st
@@ -311,6 +311,33 @@ toto'.
 	
 ]
 
+{ #category : #'tests - better parsing' }
+MicEnvironmentBlockTest >> testInputFileOnOneLineWithJunkAfter [
+
+	| mic infileReference |
+	mic := parser parse: EnvironmentOpeningBlockMarkup, 'inputFile|path=chapters/withStyle.mic', EnvironmentClosingBlockMarkup, String space, 'lkjkljklj'.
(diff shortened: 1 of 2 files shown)


kasperosterbye:
It seems like this issue should be handled in `MicInputfileBlock` rather than `MicEnvironmentBlock`. 
In `MicInputFile` this could be done as:
```Smalltalk
path: aPath
	arguments at: #'path' put: aPath trim "trim inserted".
```

The proposed solution in `MicInvironmentBlock` removes trailing blanks from any argument being the last. I cannot see any case where it would matter, but I see no reason to make the restriction that the last argument can not end in a blank, where as an argument in the middle of an argument list can.

Ducasse:
Ths problem is <! jjlkjkljkjlkj lkj !> kjlkjklj
- So why it would make sense to have an argument ending with a blank?
- And right now the parser says that the argument is jjlkjkljkjlkj lkj !> kjlkjklj while it should be jjlkjkljkjlkj lkj
Notice that the space inside are kept. 

So just doing what you suggest does not work because the logic of termination was dull.

Ducasse:
BTW I would also say that an argument in the middle should not end with a space. 
This is a recipe for bugs

Ducasse:
BTW if we accept (as it is now) to have `<!environment .... !>` on the same line
then it should not break when we have `<!environment .... !>` blblblbl
May be environment should not allowed `<!environment .... !>` on the same line

kasperosterbye:
See my comment in #521. I think this PR is a step in the right direction.

PR #526, opened 2022-08-02
Change done to the method fromUri: of MicResourceFileReference

DelGaylord:
Following our try to convert full book we saw that from Microdown to LatTeX the file:// added when we have an url caused to see the path and not the figure. @kasperosterbye i let you tell me if this is correct or not also i had revert my precedent commit and make sure to have all the files correctly so i hope i will not be like my PR yesterday

### DIFF
--- a/src/Microdown-HTMLExporter-Tests/MicHTMLExporterTest.class.st
+++ b/src/Microdown-HTMLExporter-Tests/MicHTMLExporterTest.class.st
@@ -72,7 +72,7 @@ MicHTMLExporterTest >> testAnchor [
 MicHTMLExporterTest >> testCodeCreatesInnerText [
 
 	| code |
-	code := (self parser parse: '```language=a b```') children first.
+	code := (self parser parse: '```a b```') children first.
 	self assert: code class equals: MicCodeBlock.
 	self assert: code language equals: 'a b```'
 ]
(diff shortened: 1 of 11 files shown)


kasperosterbye:
Sorry to be so critical @DelGaylord, but t is very hard for me to comment on this as you are not solving an issue - just proposing a change, so I have to guess what is your problem. 
In this case I cannot find out what your problem is.

My impression so far has been that a lot of assumptions is hidden in the Latex code and pillar that the only thing in the world is files, where as Microdown resource references handles http, https, github, comments, source and more to come.

Without a propper stated issue, I cannot judge if your problem is a problem in Microdown core, or in your Latex handling.

So please, write an issue, with an example of what in the core (here `MicResourceReference`) causes the problem, and explain to me why you think it is not a problem in the Latex code.

DelGaylord:
You don't have to be sorry it's my fault to not be accurate in my explanations. 

So to explain that better; Stéphane and I had try to convert full book with the inputfile but, at this moment we had encounters some problems. One of them was that when we checked the pdf that we get after, the figures was not there but the link at the place. It look like a square with in the middle the path and when we deleted the beginning of the path ( so the "file://") it worked. 

But i was wondering if the change i proposed was correct to do at this point or if we needed to do it somewhere else. Because i was wondering if this would happen in another convertion than LaTeX that use path.

Im sorry i can't show screenshot of that because for the moment, it's not possible for me to install pillar on my computer and with Stéphane we don't know why that and these convertions was not done on my computer.

kasperosterbye:
I briefly looked at the code for `MicLatexWriter>>visitFigure:`. The actual emit is in `includeGraphicsFor:`. 
I believe the correct code to write there is:
```smalltalk
includeGraphicsFor: aFigure

	canvas command
		name: 'includegraphics';
		optParameter: [ self writeGraphicsParametersOfExternalLink: aFigure ];
		parameter: [ canvas raw: aFigure reference path  ]
```
That is - to use the resolved reference of Figure instead of a text string.

PR #528, opened 2022-08-04
New feature MicAgenda

DelGaylord:
The test are not green because i changed several times the template used in this new feature for the creation of an Agenda and the use of MustacheTemplate. But for now is just to let know what i have done for the moment. @Ducasse


Ducasse:
Thanks Gaylord. For now I want to finish the file:// and integrate it.

PR #534, opened 2022-08-10
Fixes #533 microdown visitors are unstable

### LINKED ISSUE #533: Microdown visitors are unstable
Assume we make a new extension block - say for columns `<!columns....!>`.
Assume the columns block implements accept by sending `visitColumns:` to the visitor.

Existing visitors (there are currently 16 different visitors) will not have implemented this method.
It is up to the implementor of the columns block to remember to add an extension method to `MicrodownVisitor` to prevent a DNU if one apply a document with columns to a visitor which do not know it.

I suggest a test to make sure that all defined visit methods in all subclasses are defined in `MicrodownVisitor`.

kasperosterbye:
Added a test, and fixed the errors it found

### DIFF
--- a/src/Microdown/MicDocumentListBlock.class.st
+++ b/src/Microdown/MicDocumentListBlock.class.st
@@ -12,7 +12,7 @@ MicDocumentListBlock class >> tag [
 { #category : #visiting }
 MicDocumentListBlock >> accept: aVisitor [
 
-	^ aVisitor visitDocListAnnotation: self
+	"^ aVisitor visitDocListAnnotation: self"
 ]
 
 { #category : #visiting }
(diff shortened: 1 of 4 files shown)


src/Microdown/MicDocumentListBlock.class.st
```diff
@@ -12,7 +12,7 @@ MicDocumentListBlock class >> tag [
 { #category : #visiting }
 MicDocumentListBlock >> accept: aVisitor [
 
-	^ aVisitor visitDocListAnnotation: self
+	"^ aVisitor visitDocListAnnotation: self"
```
kasperosterbye, line 15: This visit method is not defined anywhere, so I commented it out for now

kasperosterbye:
This one has changes to the core - please take a look @Ducasse 
Btw, the CI ran this time ?!

Ducasse:
I looks ok to me. And I'm amazed that it actually run the tests!
Excellent.

PR #535, opened 2022-08-11
Optimization of the generated calendar, now we need to have to template on local

DelGaylord:
but we generate a full HTML for the calendar + need to fix the test about the use of the template


DelGaylord:
@Ducasse can you accept this PR ? It's for Christophe because he need it in order to test how works MicAgenda ^^

Ducasse:
Done :)

PR #542, opened 2022-08-17
Document browser development

kasperosterbye:


kasperosterbye:
There are a some changes to the core classes hidden in this, so it need a review. @Ducasse

Ducasse:
Sounds good to me.

PR #545, opened 2022-08-20
Cleanup package comment resource

kasperosterbye:
It now follows the same hierarchy as the packages and tags in the browser


kasperosterbye:
Internal to DocumentBrowser

PR #548, opened 2022-08-21
Adding some methods from MicAgendaGenerator to MicrodownVisitor in order to get all the tests green

DelGaylord:
I have made these change next to the comments of Kasper because i made a PR but i had not checked correctly the tests before doing it. Now all the tests are greens with the packages Microdown-Agenda and Microdown-Agenda-Tests.


Ducasse:
Ok now we should package these methods with the calendar package but we can do that in another PR.

PR #552, opened 2022-08-31
Fixed the codeblock eval=true feature + better tests

Ducasse:


src/Microdown-DocumentBrowser/MicIntermediateBlock.class.st
```diff
@@ -1,6 +1,10 @@
 "
 I am a block that is used to group other nodes internally in a microdown trees.
-I am not used by the parser, but by post-processing tools.
+I am not used by the parser, but by post-processing tools. hjkhjhkjhkjhaksjdhakjdhaskjhakfjh\
```
kasperosterbye, line 3: What is this?

Ducasse: This is nothing apparently I modified the class comments demonstrating the documentation browser when editing a class comment.

Ducasse:
Now I'm trying but failing to provide access to a field that would point to a builder in the evaluated body of the code block (similar to the stream one) but it fails so I do not get why but the change of this PR is already fixing the evaluator's code.

PR #559, opened 2022-09-28
Fixes #556, allowing tabs to be preserved in code blocks (and elsewhere)

### LINKED ISSUE #556: RichTextComposer should handle tabs in codeblock better
At the moment tabs seems to be ignored, and code is wrongly indented

kasperosterbye:
Notice that this change impacted a number of tests in latex and html. Before it was possible to write nested lists using tabs rather than space, not anymore. 
Personally I think it was a good idea anyways as it is easier to see that the indentation of nested lists fits the rules of the parser.


Ducasse:
I will have to check the books I have converted but I will check that once we release

PR #563, opened 2022-09-29
Fixes #554 microdown package rename

### LINKED ISSUE #554: Microdown package rename
I propose we rename package "Microdown" to "Microdown-Core". "Core" is used in many other multi-package projects (Bacon-Core, Clap-Core, Commander-Core...).

The tag "Core" could then be renamed "Base", (Microdown-Core-Base) which is also what is done in many other "Core"-packages (Fuel-Core-Base, FileSystem-Core-Base, Gopher-Core-Base...).

Finally, the tag "Extensions" should be renamed to "BlockExtensions" (see issue #511)

kasperosterbye:
Renames package 'Microdown' to 'Microdown-Core' as proposed in #554. 
The low-level practical aspect is that it brings 'Microdown-Core' and 'Microdown-Core-Tests' packages next to each other in the browser, but also that a 'Core' package seems to be the norm (as mentioned in #554)


kasperosterbye:
OK, this need a different approach. I think this is a do not fix what is not broken

kasperosterbye:
Will just give it an other go with the directories renamed manually

kasperosterbye:
This is a do not fix what is not broken... Not worth the effort

Issue #570, opened 2022-10-02
Fix for (bad resolution when compiling separated file) https://github.com/pillar-markup/pillar/issues/574


Ducasse:
https://github.com/pillar-markup/pillar/issues/574

```
MicRootBlock >> transformDocumentFor: aPRPDFDocument 
	"This is a hook to be able to have framework specific transformation hooks.
	see PRAbstractOutputDocument>>#buildOn:"
	
	^ aPRPDFDocument transformDocument: (self resolveFrom: aPRPDFDocument file file asMicResourceReference )
```

Ducasse:
With a nice comment

transformDocumentFor: aPRPDFDocument 
	"This is a hook to be able to have framework specific transformation hooks.
	see PRAbstractOutputDocument>>#buildOn:"
	"Note that the resolution is made against the currently transformed file and not the base directory of the project else we may lose the fact that a figure is within multiple folders within the base project. 
	Compare 
		/base and
		/base/Chapters/Chapter1/FileContainingAReferenceToFigures.md
	When resolving figures/myPng referenced from within FileContainingAReferenceToFigures.md 
	we should get /base/Chapters/Chapter1/figures/myPng!
	Using base we lose Chapters/Chapter1 and only get /base/figures/myPng.
	"
	^ aPRPDFDocument transformDocument: (self resolveFrom: aPRPDFDocument file file asMicResourceReference )

Ducasse:
fixed by https://github.com/pillar-markup/pillar/commit/85f4945f702b9a19a3423cf4bd9fa6d1d802ab6a

PR #571, opened 2022-10-02
Fix: #570 and  https://github.com/pillar-markup/pillar/issues/574

Ducasse:
Next write a nice test!


Ducasse:
I will merge it and then move it to pillar packaged with its tests.

PR #580, opened 2022-10-03
Fixes #567, MicMathInlineBlock comment does not render

### LINKED ISSUE #567: MicMathInlineBlock comment does not render
Strange but it looks like it is broken.

kasperosterbye:
This fix seems to patch the inline parser. 
I did not load the full microdown at this one, so full tests pending

### DIFF
--- a/src/Microdown-Tests/MicrodownInlineParserTest.class.st
+++ b/src/Microdown-Tests/MicrodownInlineParserTest.class.st
@@ -414,6 +414,14 @@ MicrodownInlineParserTest >> testMonospacesWithDoubleBackSlash [
 ]
 
 { #category : #'tests - format' }
+MicrodownInlineParserTest >> testMonospacesWithTwoBackslashes [
+ 	| elements |
+ 	elements := self parseAndReturnElementsOfParagraphFor: '`\a+\b`'.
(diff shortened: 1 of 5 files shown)


kasperosterbye:
Yay! 
I tried to rewrite the inline parser, but a lot of the tests depend on the internals to make it feasible.
@Ducasse - review at your leisure
In particular it is symptomatic that monospace requires a unique text child with the same substring as itself

Ducasse:
Thanks I will check. 
First I want to clean the baseline.
I'm making progress there.

kasperosterbye:
That is a difficult one, as there is some aspects of the inline parser and the inline nodes which is not needed and is redundant. But those same aspects are used as shortcuts a huge number of tests and in the visitors.
For example, the monochrome has no internal structure (is passed as raw). Except, it does. After parsing a TextNode with a **copy** of the body is created and added as a child.
And some of the external visitors depend on this fact in their implementation.
So, not only is the inline parser cumbersome, but its problems has spread randomly through the system.

Ducasse:
argh. 
shit! I do not see the problem with the copy but I trust you on that.
So I suggest that we release first.

PR #588, opened 2022-10-05
This patches the error of rubrics

kasperosterbye:
It was a known issue which surfaced in a new way.
There will now ne rendered a space before and after an image when rendered asText.


kasperosterbye:
It is vital that this does not break, even if it is not the right way to handle it

PR #593, opened 2022-10-10
Adding a newLine before each table when converted from Microdown to LaTeX

DelGaylord:
While converting books from Pillar to Microdown, I also checked that we get the same result with the Microdown file than the Pillar file. But there was a problem with the table which was put next to the text before them. So I added a newLine in order to get the good result when we get the LaTeX file to pdf.


Ducasse:
Tx gaylord

PR #607, opened 2022-10-15
FIxes 543 nice environment

kasperosterbye:
Made a test similar to original proposal.
Implemented as suggested in the issue.
Fixed a number of tests that depended on near empty lines with just a single tab in the line.


Ducasse:
I think that this is good that the blocs can be tabbed only if they are nested. So it means that people will not notice it instead they use it.

PR #626, opened 2022-11-03
MERGE ME - Inline rewrite all green

kasperosterbye:


Ducasse:
Thanks for the update. 
I'm chasing a bug and I should do one xp for a student and I will have a look. So probably in the evening.

kasperosterbye:
@Ducasse 
There are two different kinds of inconsistencies with the old inline parser and old inline parser blocks

1. The previous blocks had several methods for accessing the original source code. That was used in two different situations. a) As a lazy way to do tests (avoiding to dive into the AST for validating), and b) several places used those source methods mistakingly (missing visits to children). I have in general changed both a) and b) to behave correctly.

2. The new parser has a different algorithm for parsing. Hence Macrodown which extended the old algorithm will not work. It is the only case I know of where the inline parser algorithm was extended. I will not fix this.

I tried Agenda, which works fine. 
I **do not expect** this to break your book making, but I obviously have not tried. Latex translation might have a issues in category 1. above.

I will not do anything more on this one now, I considerer it working except for the risk of a few issues in category 1 which so far has been easy for me to fix.

Ducasse:
Ok I will check. I was removing root trees today.... and now meeting with the town. 
but I'm reducing my todo list.

Ducasse:
Hi kasper 
Why did you close this PR?
This was still on my todo list

kasperosterbye:
OK, I really want your input on this. It is massive and invasive.

PR #648, opened 2023-03-08
Some styles were updated in Pharo. This updates Microdown

jecisc:


jecisc:
I guess it fails because Pharo needs time to bootstrap a new image with the changes

MarcusDenker (approved):

jecisc:
I added that we build on P11

demarey (approved):

PR #652, opened 2023-04-21
Add codeblock evaluation extension

estebanlm:
I added an extension that shows an "executable" code block, so I can add executable statements in the new welcome browser.
(I do not like it completely, but more or less :P)

Screenshot from 2023-04-21 18-16-32


estebanlm:
@Ducasse ping :)

PR #654, opened 2023-05-09
Remove dead method

jecisc:
I removed a method that is not used and that uses an API that will be deprecated.


jecisc:
Apparently this is used but not in the code loaded in Pharo. I'll do another one to update the API used

PR #675, opened 2024-01-18
Add missing method to render package comments

hernanmd:
Reported in https://github.com/pharo-project/pharo/issues/15916

This PR just add to Pharo 12 a method for rendering package comments, which was lost from Pharo 11.


Ducasse:
I will close it because I do not want such extension in Microdown it should not be there but in BeautifulComment
#684

PR #678, opened 2024-01-27
676 use is not nil instead of not nil

### LINKED ISSUE #676: Use isNotNil instead of notNil
Use isNotNil instead of notNil also in Microdown for Pharo 12

astares:
Fix #676


Ducasse:
I will check. My problem is that the development of Microdown must not be in P12 else I'm dead and cannot continue to develop Pillar and Microdown. 
So for now I will just ignore Pharo12 and Pharo12 can just load a tag version for example v2.4.2 (that I will do with the removal of buildMicroDownUsing moved to Beautiful)

astares:
???

Ducasse:
I will check apparently guille fixed OSSProcess so I will be able to move the dev to P12. It will make my life a lot easier.

astares:
Ideally we have "Pharo12" or later "Pharo13" branches on each project that Pharo uses/includes in the standard image 
and load them during the development iteration. 

This makes providing and merging PR's a lot easier. Also this way one can more easily backport changes also and it is more obvious also to contributors.  Would be a little bit similar to "open editions" back in ENVY.

When Pharo then later wants to make the regular yearly final release each project can tag a depicted tagged version then also additionally. 

If (as it is now) every project works differently we just increase efforts and confusion. Using the above proposed approach would require no additional tools - it just requires only a little bit more discipline and consistency.

Ducasse:
Yes except that Pillar did not work in P12 so there was no way that I could work in P12. So now Guille helped me and I will migrate and release a version of Pillar for P12.
And in Pillar I do not want to have a floating microdown version because I do not release pillar based on Pharo release
I release when I can and a release should only use fixed versions of all its components.

Ducasse:
I will close this issue and basically redo it.

astares:
yes, because only a few cases are left now in P12 to convert

PR #680, opened 2024-01-27
Do not use deprecated #next of UUIDGenerator in MicHTMLDocument>>#tempFileName

### LINKED ISSUE #679: Do not use deprecated #next of UUIDGenerator in MicHTMLDocument>>#tempFileName
In MicHTMLDocument>>#tempFileName

we should use

```Smalltalk
UUID new asString36
```

instead of 

```Smalltalk
UUIDGenerator new next asString36
```

for Pharo 12

astares:
Fix #679


Ducasse:
I will integrate it as soon as I succeed to load MicrodownDev in P11.

Ducasse:
In Pharo11 the method is not deprecated. And right now I cannot move to P12 because Pillar does not work in P12 :(. 
The HTML package should not be loaded for the Pharo release so I will close the issue and see if one day I migrate to P12.

PR #683, opened 2024-02-02
Move extension to renamed Package

guillep:
RPackage was renamed Package in Pharo 12. Fix extension method

### DIFF
--- a/src/Microdown/Package.extension.st
+++ b/src/Microdown/Package.extension.st
@@ -1,7 +1,7 @@
-Extension { #name : 'RPackage' }
+Extension { #name : 'Package' }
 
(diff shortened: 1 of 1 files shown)


jecisc (approved):

Ducasse:
I want to move all these extensions to Beautiful comments I have to see how I can do that #684.
Now I cannot load Microdown in P11 because of an extension and I cannot run Pillar in P12 because of OSSprocess 
So I will concentrate on P11 because I want to be able to produce book and slides. 
For now this is key for me.

Ducasse:
I will close it and remove all the build* methods. I moved them to BeautifulComments dev (not master because I do not understand the conflict).

PR #737, opened 2024-05-14
Correction and update of Microdown-Blog

moufort:
I had fix test, add method for generate a List with all month since January 2014.
And I have fix the method firstParagraphOfBlock for take a number of word in the paragraph


Ducasse:
Hello Quentin it looks good. Now you should have started from the integrated version I merged because your PR contains all the  previous changes. 
I will merge but it means that once it is merged you should pull from master and create another branch for your next iteration. else you will republish again all your code. 
I will add your packages to the baseline so like that your tests will be executed too.

PR #740, opened 2024-05-15
add link and test in Web developpemment

moufort:
I had a link for the header of file in HTML and for each date.
I had a ellipsis to paragraph block where there too long.

### DIFF
--- a/src/Microdown-Blog-Tests/MicSummarizerTest.class.st
+++ b/src/Microdown-Blog-Tests/MicSummarizerTest.class.st
@@ -94,6 +94,153 @@ When you participate to the mooc you get access to the quizz and the credit vali
 	^ file
 ]
 
+{ #category : 'as yet unclassified' }
+MicSummarizerTest >> htmlBodyTest [
+
+	^ '<body><main role="main">
+<h1><a target="_blank" href="''''/''''">A Cool Story</a></h1>
+<p>Pharo is cool but _this is_ a superlong _paragraph_ Simple powerful language: No constructors, no ...</p>
+<h1><a target="_blank" href="''''/''''">Pharo is cool</a></h1>
+<p>If you are either a beginner or an expert in object-oriented programming, _this MOOC_ will ...</p>
+<h1><a target="_blank" href="''''/''''">Mooc Pharo</a></h1>
+<p>Welcome to the Pharo Mooc (a set of videos, exercises, challenges, and miniprojects).</p>
+<ul>
+<li><a target="_blank" href="http://pharo.org/">January 2014</a></li>
+<li><a target="_blank" href="http://pharo.org/">February 2014</a></li>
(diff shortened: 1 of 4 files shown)


Ducasse:
Salut quentin 
regarde mes remarques.

src/Microdown-Blog-Tests/MicSummarizerTest.class.st
```diff
@@ -94,6 +94,153 @@ When you participate to the mooc you get access to the quizz and the credit vali
 	^ file
 ]
 
+{ #category : 'as yet unclassified' }
+MicSummarizerTest >> htmlBodyTest [
+
```
Ducasse, line 99: for the single summariser I do not get why we would have to have all the links.

src/Microdown-Blog/MicSingleSummarizer.class.st
```diff
@@ -2,12 +2,63 @@ Class {
...
 	#category : 'Microdown-Blog',
 	#package : 'Microdown-Blog'
 }
 
+{ #category : 'rendering' }
+MicSingleSummarizer >> createHtmlFile: aMicRoot [
+
+	| a q date header |
+	a := (MicHTMLVisitor new visit: aMicRoot) at: 1.
+
```
Ducasse, line 18: we can add on microroot something like 
```
date
     self children first date
```

```
MetaData >> date
     ....
     with a default date if there is none in the metadata

```

this will probably only works when we have metadata so on MicElement we should define another 

```
date
    ^ Date ....1980
```

src/Microdown-Blog/MicSingleSummarizer.class.st
```diff
@@ -2,12 +2,63 @@ Class {
...
+	^ destinationDirectory
+]
+
+{ #category : 'accessing' }
+MicSingleSummarizer >> destinationDirectory: aDestination [
+
+	destinationDirectory := aDestination
+]
+
```
Ducasse, line 49: indeed the summariser as well as all the components date sorter, blogentry...
should all work from a directory to a directory. 

So I asked myself what is filePath vs. destinaryDirectory: why this is not a sourceDirectory

src/Microdown-Blog/MicSingleSummarizer.class.st
```diff
@@ -88,3 +148,17 @@ MicSingleSummarizer >> summarizeFile: aFileReference [
...
 	^ self summarize: p
 ]
+
+{ #category : 'parsing' }
+MicSingleSummarizer >> transformHeaderToLinkHeader: aHeader [
+
+	| headerLink |
+	headerLink := MicHeaderBlock new.
+	headerLink
+		addChild:
```
Ducasse, line 158: You can extract this expression in makeAlink: to:

PR #741, opened 2024-05-17
Fix deprecated call

jecisc:
Fix a call to a deprecated class to be removed in P13


Ducasse:
for now we are developing in P12 but I merged your PR in P13 branch. 
Thanks

jecisc:
This can also work in P12. 
I'm fixing P13 in priority because I want to unload the code we deprecated during P12 development

PR #742, opened 2024-05-17
update Microdown-Blog

moufort:
I add test.
I add MicBlogCreator who can create an tree of HTML file in a target repository with a source repository


Ducasse:
sound good

PR #743, opened 2024-05-17
Mic parent child checker

melkiyasser:
Parent Child Checker repaired , make sure each parent/Child point to each other correctly 
will flag a well formed document  that contains italic / bold as broken because the leaf text elements in such document have a nil parent

### DIFF
--- a/src/Microdown/Microdown.class.st
+++ b/src/Microdown/Microdown.class.st
@@ -147,6 +147,16 @@ Microdown >> parse: aStreamOrString [
 ]
 
 { #category : 'facade' }
+Microdown >> parseFile: aFile [
+		
+		|root|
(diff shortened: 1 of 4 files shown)


src/Microdown/Microdown.class.st
```diff
@@ -146,6 +146,16 @@ Microdown >> parse: aStreamOrString [
 	^ MicrodownParser parse: aStreamOrString
 ]
 
+{ #category : 'facade' }
+Microdown >> parseFile: aFile [
```
Ducasse, line 156: Why this is in this PR?

Ducasse:
I will integrate the changes but there were messy.
You should create clean branches

PR #745, opened 2024-05-21
Add test to Microdown blog

moufort:
I add test


Ducasse:
@Moufort the following test is failing
testCreateAllHtmlFile

Can you have a look?

PR #759, opened 2024-05-31
Microdown url references to web locations

### LINKED ISSUE #16512: Microdown URL references to web locations requiring authentication fail and do not open a browser
When within any Microdown handled text (for instance in a class comment) one creates an anchor link to a web 
resource that is behind an authentication layer the system gives an error instead of opening the browser as for 
other accessible URLs.

For instance if you point in a class comment to a private repo in github https://github.com/myuser/myprivaterepo
the link will not open but Pharo will present the debugger with a **MicResourceReferenceError**.

image

To reproduce in Pharo 12:
- create a package "Foo"
- create a class "Bar"
- in the class comment add the following comment
```
This is a class from [https://github.com/myuser/myprivaterepo](https://github.com/myuser/myprivaterepo)
```
image

- save and accept the class comment
- now click on the link in the class comment

image

=> you get the exception

The expectation is that Pharo also for such URLs just opens the external webbrowser

matburnx:
Fix https://github.com/pharo-project/pharo/issues/16512
Now it always opens the web browser with the URL given, even if it's inaccessible.
Though, the user will only see that the URL is not working once the browser has loaded the page.


Ducasse:
Hello I do not get why the GH has been changed to run on P13. 
The development of Microdown is using P12.

matburnx:
Hello, I made a mistake when linking my Pharo image to my repo.
I have now a branch that only has my commit, but I think it would need to close this PR and open a new one instead of just changing this one.

matburnx:
The new and corrected one is here https://github.com/pillar-markup/Microdown/pull/766
So I'll close this one

Ducasse:
Thanks mateo.

PR #766, opened 2024-06-03
Microdown url references to web locations

### LINKED ISSUE #16512: Microdown URL references to web locations requiring authentication fail and do not open a browser

matburnx:
Fix https://github.com/pharo-project/pharo/issues/16512
Now it always opens the web browser with the URL given, even if it's inaccessible.
Though, the user will only see that the URL is not working once the browser has loaded the page.
Correction of https://github.com/pillar-markup/Microdown/pull/759

### DIFF
--- a/src/Microdown/MicHTTPResourceReference.class.st
+++ b/src/Microdown/MicHTTPResourceReference.class.st
@@ -39,8 +39,7 @@ MicHTTPResourceReference class >> getBytesByHttp: uri [
 		get.
 		^ client entity bytes ] 
 	on: ZnHttpUnsuccessful, NetworkError, MessageNotUnderstood
-	do: [ :error | 
-			error resignalAs: (MicResourceReferenceError new messageText: 'Could not access ', uri printString)  ].
+	do: [ WebBrowser openOn: uri ].
 ]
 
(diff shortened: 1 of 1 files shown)


Ducasse:
I do not get it I thought that I already integrated a similar fix.

Ducasse:
But no it was.

Ducasse:
I will revert this changes because it breaks some tests

PR #767, opened 2024-06-05
MicrodownBookTester

melkiyasser:
make sure code blocs with code examples  are correct


Ducasse:
I integrate this and I will rename some classes.

PR #768, opened 2024-06-06
BookTester Tests

melkiyasser:
Fixed tests for bookTester for code bloc example


Ducasse:
Yasser in microdown the package for tests are named with Tests at the end. 
So the test package  is Microdown-BookTester-Tests and not Microdown-Tests-BookTester
Can you update this?

PR #769, opened 2024-06-06
Refresh class comments in Microdown package

Gabriel-Darbord:
In my opinion, class comments are especially important for the Microdown model, parser, and manifest, as they showcase the features of the project.

The main goal was to fix the feature showcases that didn't display correctly.
I took the opportunity to fix typos and grammar, but I didn't update (much) if I didn't fully understand the meaning.

For example, here's the end of the manifest comment before:
image
After:
image


Ducasse:
Merci gabriel!!!

PR #794, opened 2024-07-21
Added a first test in full without beamer generation.

Ducasse:
getting started


Ducasse:
This is strange because all the tests are green on my machine. So may be this is a end of line problem.

PR #796, opened 2024-07-21
try to fix the issue 771 by replace originalString by fullname

moufort:


Ducasse:
Hi 

Can you define 
fullNameString in the MicInputFileBlock class and use it?
instead `anInputfile path fullName` -> anInputfile fullNameString

PR #801, opened 2024-07-25
Issue 698 new optimization of MicInlineDelimiter

moufort:
I did a new optimisation of Regex of MicInlineDelimiter.
When I use the command: [ MicInlineDelimiter allRegex ] bench
I obtain over 700 million of iteration in 5 second, against only 70 000 with the previous version.


Ducasse:
Nice. This is not everyday that we get such speed up (even if Pharo is fast).

PR #802, opened 2024-07-29
Issue 780 css framework

moufort:
We add all css framework in a local folder ressource.
Now, we can load the css file from the local repository instead of internet


moufort:
That's work on my computer but not here.
I think that have a problem with the path of css file in the github.

PR #805, opened 2024-07-30
issue 804 problem running build server

moufort:
Try to fix the path problem by changing FileSystem workingDirectory by FileLocator imageDirectory


Ducasse:
It did not work. I will check that Guille in the future

PR #808, opened 2024-08-01
issue 785  Better file handling error

moufort:
improve the error sentence of MicFileRessource when we don't find the file

### DIFF
--- a/src/Microdown-Tests/MicFileResourceReferenceTest.class.st
+++ b/src/Microdown-Tests/MicFileResourceReferenceTest.class.st
@@ -277,7 +277,7 @@ MicFileResourceReferenceTest >> testLoadNotFoundMicrodown [
 		ref loadMicrodown.
 		self fail
 	] on: Error do: [ :e |
-		self assert: e messageText equals: 'Microdown file not found' ]
+		self assert: e messageText equals: 'The Microdown file at "/readme.md" are not found' ]
 ]
 
 { #category : 'tests' }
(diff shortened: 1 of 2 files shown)


src/Microdown-Tests/MicFileResourceReferenceTest.class.st
```diff
@@ -277,7 +277,7 @@ MicFileResourceReferenceTest >> testLoadNotFoundMicrodown [
 		ref loadMicrodown.
 		self fail
 	] on: Error do: [ :e |
-		self assert: e messageText equals: 'Microdown file not found' ]
+		self assert: e messageText equals: 'The Microdown file at "/readme.md" are not found' ]
```
Ducasse: is not found.

src/Microdown/MicFileResourceReference.class.st
```diff
@@ -63,7 +63,7 @@ MicFileResourceReference >> contents [
 	[^ self fileReference isDirectory ifTrue: [''] ifFalse: [self fileReference contents] ]
 	on: FileDoesNotExistException 
 	do: [ :error | 
-		error resignalAs: (MicResourceReferenceError new messageText: 'Microdown file not found')  ]
+		error resignalAs: (MicResourceReferenceError new messageText: 'The Microdown file at "', self fullName , '" are not found')  ]
```
Ducasse: May be should put it in singular.

PR #824, opened 2024-08-05
822 figuresfpng is breaking

### LINKED ISSUE #822: ![](figures/f.png) is breaking
```
![](figures/f.png)
```

Produces an error

Ducasse:
So fix #822
hasLabel usage was wrong


Ducasse:
broken tests are unrelated

PR #828, opened 2024-08-05
change cssFramework basePath to destinationPath

moufort:
try to fix the current broken test


moufort:
I closed because my commits don't fix the problem

PR #839, opened 2024-08-06
Update README.md

hernanmd:
Update load script.
Fix grammar.


Ducasse:
For now I'm still working in P12 because pillar should work. 
But thanks for the readme update

Issue #831, opened 2024-08-06
Use annotatedParagraph for Latex outputter


Ducasse:
### Task 1 
We should verify if 
 
```
!! Important This is an important information
```

or 

```
!! Important 
This is an important information
```

are actually working. Not sure about the second one.

- Do we parse correctly? do we have tests to cover the parser.
- We should rename MicAnnotatedBlock in MicAnnotatedParagraph
- Do we obtain a correct MicAnnotatedParagraph
- Introduce visitAnnotatedParagraph:
- Then we should check that the LatexExporter export correct latex

```
\begin{Important}
texte
\end{Important}
```

### Information related to the task

In the template we have the following definition

```
% highlighted paragraphs
\newenvironment{note}{%
  \begin{leftbar}\textsf{\textbf{Note}}\quad
}{\end{leftbar}}

\newenvironment{todo}{%
  \begin{leftbar}\textsf{\textbf{To do}}\quad
}{\end{leftbar}}

\newenvironment{important}{%
  \begin{leftbar}\textsf{\textbf{Important}}\quad
}{\end{leftbar}}
```

### Task 2.
A first version 

```
!!note texte
```

should generate

```
\begin{note}
texte
\end{note}
```

```
!!important texte
```

should generate

```
\begin{important}
texte
\end{important}
```

### Points to consider

If the user writes

```
!!fooooooo 
esrsrsrr
```

then we generate

```
\begin{note}
fooooooo texte
\end{note}
```

If the user writes 

```
!!imporTant texte
```

we still generate 

```
\begin{important}
fooooooo texte
\end{important}
```

- In the future we can extend it so that if the 

\newenvironment{annotatedParagraph}{%
  \begin{leftbar}\textsf{\textbf{#1}}\quad
}{\end{leftbar}}

```
!! Undefined
texte
```

generates

```
\begin{annotatedParagraph}{Undefined}
texte
\end{annotatedParagraph}
```

PR #843, opened 2024-08-08
Issue 831 use annotated paragraph for latex outputter

moufort:
Issue #831
I add test for annotatedParagraph and export it in Latex


Ducasse:
Super bien.

PR #853, opened 2024-08-11
Fixing beamer writer

Ducasse:
Fixing beamer writer


Ducasse:
May be I duplicated some tests I will have to check

Issue #845, opened 2024-08-09
More annotatedParagraph


Ducasse:
### Task 2.
A first version 

```
!!note texte
```

should generate

```
\begin{note}
texte
\end{note}
```

```
!!important texte
```

should generate

```
\begin{important}
texte
\end{important}
```

### Points to consider

If the user writes 

```
!!imporTant texte
```

we still generate 

```
\begin{important}
fooooooo texte
\end{important}
```

If the user writes 

```
!!Foooo  texte
```

we still generate 

```
\begin{note}
Foooo  texte
\end{note}
```

### For later

We can extend it so that if the 

\newenvironment{annotatedParagraph}{%
  \begin{leftbar}\textsf{\textbf{#1}}\quad
}{\end{leftbar}}

```
!! Undefined
texte
```

generates

```
\begin{annotatedParagraph}{Undefined}
texte
\end{annotatedParagraph}
```

PR #857, opened 2024-08-14
Issue 845 more annotated paragraph

moufort:
Issue #845 
I add valid label and a default label. If the label is not a valid label, then we use the default label

### DIFF
--- a/src/Microdown-HTMLExporter-Tests/MicHTMLVisitorTest.class.st
+++ b/src/Microdown-HTMLExporter-Tests/MicHTMLVisitorTest.class.st
@@ -147,14 +147,16 @@ MicHTMLVisitorTest >> testCreateAnchorWithLink [
 MicHTMLVisitorTest >> testCreateAnnotationCompound [
 
 	| result |
-	
 	result := writer convertMicString: '# This is a title
-!!LabelOfAnnotatedBlock Body of annotated block
+!!note Body of annotated block
 
 this is another string'.
-	self assert: result trimBoth equals: '<h1>This is a title</h1>', newLine ,
-'<section id="LabelOfAnnotatedBlock"><label for="LabelOfAnnotatedBlock">Body of annotated block</label></section>', newLine ,
-'<p>this is another string</p>'.
+	self
+		assert: result trimBoth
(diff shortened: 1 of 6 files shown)


src/Microdown-HTMLExporter-Tests/MicHTMLVisitorTest.class.st
```diff
@@ -147,25 +147,30 @@ MicHTMLVisitorTest >> testCreateAnchorWithLink [
...
-	
 	result := writer convertMicString: '# This is a title
-!!LabelOfAnnotatedBlock Body of annotated block
+!!note Body of annotated block
 
 this is another string'.
-	self assert: result trimBoth equals: '<h1>This is a title</h1>', newLine ,
-'<section id="LabelOfAnnotatedBlock"><label for="LabelOfAnnotatedBlock">Body of annotated block</label></section>', newLine ,
-'<p>this is another string</p>'.
+	self
```
Ducasse, line 154: To me it looks strange to create a heading 1 for a note. 
Why did you suggest this?

src/Microdown/MicAnnotatedParagraph.class.st
```diff
@@ -26,9 +29,23 @@ MicAnnotatedParagraph >> accept: aVisitor [
...
+MicAnnotatedParagraph >> defaultLabel [ 
+	
+	^ #note
+]
+
+{ #category : 'testing' }
+MicAnnotatedParagraph >> initialize [
+
+	super initialize.
+	validLabel := { #note . #important . #todo }
```
Ducasse: Could use plurals eg validLabels 
Tx

src/Microdown-HTMLExporter-Tests/MicHTMLVisitorTest.class.st
Ducasse: And we can let it for now and see later.

moufort: The test with h1 already exist when I have adding the verifiedLabels. I only correct the test for let him pass green

Ducasse:
Thanks for the explanation.

PR #858, opened 2024-08-14
Fix unreadable code in Pharo Light theme

hernanmd:
Fix an issue reported in https://github.com/pharo-project/pharo/issues/16876 that causes all Microdown Smalltalk code to be unreadable in the Calypso class comments and documentation browser.

Now Microdown code style follows the user's code style settings. This is done by resetting the textAttributes variable in the Microdown Smalltalk text styler during initialization, so it uses the user's configured code style.

Screenshot 2024-08-14 at 19 30 24


Ducasse:
Tx hernan!

hernanmd:
In the latest P13 this PR isn't integrated yet, is that expected?

PR #877, opened 2024-09-11
Enh support sync method definition

Ducasse:
- Added support for sync definition
Now to make progress (eg generating a good report) this PR should be integrated first.


Ducasse:
I will fix the test in another PR.

Issue #810, opened 2024-08-02
About duplicated anchors


Ducasse:
We should not just create a result for the last occurrences of a duplicated anchor but for all of them. Else we will not be able to say to the user that anchor1 in file1 is duplicated with anchor2 in file2.

We should write a test to check if we report two results. 

If we do not handle duplicated anchors well then the logic should be 
- add a result all the time
- then check **at the end** to remove the uniquely defined ones from the results.

Ducasse:
@melkiyasser

PR #890, opened 2024-09-29
Issue 810 About Duplicater Anchor

moufort:
Issue #810
I add a method for add to duplicated anchor and to results the first occurence of the duplicated anchor
And i add this method in the method who collect duplicated anchor


Ducasse:
Ok I have to digest the logic :)

Ducasse:
Because I was thinking to add all the anchors and check at the end.

Issue #815, opened 2024-08-03
ReferenceChecker should handled the inputFiles the same for the checkDir: than the checkProject:.


Ducasse:
Right now 

- checkProject: is looking for the transitive closure of the files via inputFile (using the fileCollector)
- checkDirectory: does not look for the transitive closure and this is not good.

both checkProject: and checkDirectory: should have the same semantics. 

We should add a test clearly showing that they is not difference once this problem is fixed.

Ducasse:
@melkiyasser

Ducasse:
checkProject: is basically looking for all the files it can reach starting from an input.  When making checkDir: using the same logic with all the files in a directory, the worklist should be shared among all the files in checkDir: to avoid to treat files multiple times (it could happens if we have several files in the directory pointing a same inputFile for example).

Ducasse:
We should check the PR of Quentin because I have some doubts.

PR #892, opened 2024-10-06
815 reference checker should handled the input files the same for the check dir than the check project

moufort:
Issue #815 
I change checkDirectory for using the same semantic as CheckProject
I also change visitRoot: of MicFileCollector for didn't visit a file already visited in the directory


src/Microdown-BookTester/MicReferenceChecker.class.st
```diff
@@ -72,8 +72,21 @@ MicReferenceChecker >> addDuplicatedFirstAnchor: anAnchor [
 { #category : 'main API' }
 MicReferenceChecker >> checkDirectory: aDir [
 	"Take the directory, parse all its children with microdown file parser and let the visitor visit each time then return visitor is ok which should be true if every thing is okay, the visitor turned out to treat the many documents that it visits as one, so if anchor is duplicated in another file it will detect that . "
-	
-	self checkList: aDir allFiles
```
Ducasse, line 76: I asked myself why we do not call checkList:
Another question I asked my self was why the collector is not shared between the checkDirectory: and checkProject:

moufort: Here I reused the checkProject: code for having the same semantic in both of checkProject and checkDirectory. chekList: is call at the end of the method.

For the collector who aren't shared between the checkDirectory: and checkProject:, I didn't think about it, maybe I can create a collector variable for the class

PR #917, opened 2024-11-04
Some cleanups

jecisc:
- Put methods in the right protocols
- Remove methods with equivalents in the superclass
- Remove method with undeclared var
- Move Error class from test to model package


hernanmd (approved):
Approved

PR #923, opened 2024-12-16
Implements compatibility with Strict Symbol Comparison feature in Pharo

carolahp:
This PR depends on Pharo [PR#14586](https://github.com/pharo-project/pharo/pull/14586) in Pharo.
This a draft PR, please do not review yet


Ducasse:
Hi caro I think that I will try to remove the reliance on Symbol and use plain string.

carolahp:
Hi Stef, that's a good idea, although for the moment we will keep the strict comparison between Strings and Symbols as an optional feature

PR #925, opened 2024-12-17
Add GitBridge to Microdown

jecisc:
With this it is possible to do:

`MicBridge root` or `MicBridge resources` to get the file references you want


Ducasse:
Oh super thanks!

PR #927, opened 2025-01-02
Proposing a simple workaround that add empty columns for short rows

### LINKED ISSUE #926: MicRichTextTable dows not accept "short rows"
```st
MicrodownPresenter new
	documentFromString: '|Table|with |3 columns|
---
|this row|has|3 columns|
|this one|has only 2|' ;
build ;
open
```

raises an error.

note: The HTML exporter works on the same example

NicolasAnquetil:
fix #926


Ducasse:
Could you add some tests covering the feature?
I would like to understand what is the problem and the feature you want.

NicolasAnquetil:
sure.
see new commit

PR #929, opened 2025-01-31
Add package tags as a semantic entity

iglosiggio:
This way we can quickly reference and navigate to a given package tag.

Made during the pharo sprint with @matijakljajic

### DIFF
--- a/src/Microdown-RichTextComposer/MicSemanticAction.class.st
+++ b/src/Microdown-RichTextComposer/MicSemanticAction.class.st
@@ -59,7 +59,7 @@ MicSemanticAction >> computeEntity [
 	"either 'Point' for a class or #'''System-Caching''' for a package"
 	size = 1 
 		ifTrue: [ self getClassOrNil
-						ifNil: [self getPackageOrNil ].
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks I will have a look. 
Would be good to have tests and getXXX is not that idiomatic of Pharo.
I will have a look.

Ducasse:
I know that the existing code is using this getX stuff.

iglosiggio:
Yes, we continued the code following what was already there. Now it's probably time to improve it.

OTOH there's no test on this PR and we are adding a new feature. That's not right :P

My rewrite would be:
- Remove the object mutation from the getX methods
- Add a field containing the list of supported "entity providers" (an entity provider is something that respondsTo #value: and expects the token list as argument)
- Make the initial value of this list something like: `{ [ :tokens | self maybeGetClass ]. [ :tokens | self maybeGetPackage ]. [ :tokens | self maybeGetTag ] }` (if we put a bit more work we can migrate the whole computeEntity method to a series of calls like that)
- Then the implementation of computeEntity is just `entity := providers detect: [ :provider | provider value: tokens ]`.

Ducasse:
I checked and it looks ok for now. 
So I will integrate it.

PR #951, opened 2025-07-07
604 micro shared pool should be removed

kasperosterbye:
The original issue was:

> It seems like a left over from a renaming.
> 
> Btw, why is its superclass called MicMicrodownSharedPool??
> It would seem more natural to be MicParserSharedPool, or MicDelimitersSharedPool

The class name is today MicSharedPool, and the two subclasses were empty, and had no references.

So this just removes those two classes (and is an attempt for me to write a response again.


Ducasse:
I'm not sure that I want to work on P14 for now. 
So I will redo the changes in P13.
Release Microdown and Pillar for P13 then I will come back to the gitbridge issue mixed with the removal of the classes.

PR #953, opened 2025-07-10
952 table row ending in  with no character after the last 

kasperosterbye:
I fixed an error in the tables ending with "|" or with "|  ". And added a test class for that


Ducasse:
I harvested the move of gitbridge to Microdown-Blog-Tests.
Now I will 

I do not think that dev (and the new release should be pushed Pharo 13) so once I release Microdown 2.7.3 I will not merge it in Pharo 13 before people suggest it.

Ducasse:
I redid the changes and push a commit. All the tests are green!

PR #954, opened 2025-07-12
834  {} is not parse well 

kasperosterbye:
This adresses and solved the issues from issue 834. And only that.


kasperosterbye:
The key solution is to add the ifNil: in
closeMe 
	super closeMe.
	body ifNil:[ body := ''].
	body := [ STONJSON fromString: '{', body, '}' ] on: Error do: [ :ex |
		| dict |
		bogusParsing := true.
		dict := Dictionary new at: self keyForUnparsableContents put: body; yourself ]

Ducasse:
I integrated the changes in dev. So I close it. Tx!
Now ready for the release.

PR #972, opened 2025-08-29
Merge dev in P14

jecisc:


Ducasse:
Oops I did not see this and I will need help to understand

PR #981, opened 2025-11-17
Use less obscure way to reset delimiters of MicInlineDelimiter

jecisc:
We might deprecate #noteCompilationOf:meta: with Guille since it's just an alternative way to react to method changes and we already have the announcements


jecisc:
Hum... In fact it does not do what was done before. I'll close it for now

Issue #618, opened 2022-10-27
HTMLVisitor loops endlessly


bourahma:
````

MicHTMLVisitor 
	serializeToHTMLDoc:
(Microdown parse: '# hello world

bllbl

# section2 

titit
')

````

kasperosterbye:
What do you mean?
If I run your example I get an error because MicHTMLVisitor do not provide a default CSS, but it does not loop endlessly.

PR #983, opened 2025-12-17
Allow MicCSSEntity to overwrite existing file

### LINKED ISSUE #982: MicHTMLVisitor examples cause an infinite loop
To reproduce:
Run two of the following class-side examples, or the same example twice:
* `MicHTMLVisitor class>>#exampleHTMLDocument`
* `MicHTMLVisitor class>>#exampleHTMLDocumentWithFigure`
* `MicHTMLVisitor class>>#exampleHTMLDocumentWithLink`

What did I expect:
The given example HTML and CSS is output to the `output` directory in `FileSystem workingDirectory`.

What happens:
On the second run there is an infinite loop.

AndrzejProchyra:
When telling its CSS file to copy itself to a destination directory, MicCSSEntity was giving its MicCSSFile the full path, including the file basename. If the file alredy existed at that path, the MicCSSFile was attempting to delete the full path and appending the basename again, leading to an infinite loop.

This completely moves the responsibility for knowing the basename to MicCSSFile.

A proposal to fix #982 and #618


Ducasse:
Thanks a lot andrzej!

Issue #980, opened 2025-11-06
MicFileResourceReference >> fromFileRef: answers invalid Windows full name


Driolar:
Doing something like
```
(MicFileResourceReference fromFileRef: 
'C:/cygwin64/home/driolar/Booklet-PharoGraphs-master/Chapters/Chapter1/figures/Paris_Metro_map_1024.pdf' asFileReference) fullName 
```
leads to

*>>> '/C:/cygwin64/home/driolar/Booklet-PharoGraphs-master/Chapters/Chapter1/figures/Paris_Metro_map_1024.pdf'*.

Note the leading slash before the drive letter that makes the path invalid on Windows.
As a consequence, TeX cannot find the figures when building a PDF on Windows.

The leading slash is set in:
```
MicAbsoluteResourceReference >> path
	"return the path part of my uri"
	^ '/', uri path
```

kasperosterbye:
In essence there are two directions for a solution.  a) Check the OS, and fix the error for windows
b) Check if the first part of the file is "X:", and then assume it is the full file name  There is no test method for this issue. And I would not be able to write one for a) as I am on a MacOS. So I would prefer b). "X:" is not allowed on MacOS, but is legal filename on Linux, so if someone made a root directory named "X:", it can not be sure if it is actually a directory name, or the root of a file name.

Driolar:
I am on Windows, so I could provide the variant a) solution if you agree.

kasperosterbye:
Yes, a solution is often better than a discussion :-), so under that offer I think a) is perfect

PR #993, opened 2026-02-10
Fixed MicAbsoluteResourceReference path in case of Windows drive letter leading.

Driolar:
This is to solve issue #980.

### DIFF
--- a/src/Microdown-Tests/MicrodownTest.class.st
+++ b/src/Microdown-Tests/MicrodownTest.class.st
@@ -66,6 +66,15 @@ Pharo is cool
 ' ].
 ]
 
+{ #category : 'tests' }
+MicrodownTest >> testAbsoluteWindowsPath [
+
+	| absoluteReference |
+	absoluteReference := MicAbsoluteResourceReference newFromUri:
+		                     (ZnUrl fromString: 'c:/temp/foo.txt').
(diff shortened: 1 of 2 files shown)


kasperosterbye:
I like the structure og this. It explicitly checks if you are on windows, and if you are it checks if the second character is a $:
That the text fails is due to something else (and is an other issue, which is that OS should by default be as it is now, but then one should in some cases (like in test) say "we are now on windows, or we are now on Linux".

Driolar:
Running all Microdown tests with Dr. Tests, 17 errors (all in `MicTemplatedWriterTest`) and 19 failures result. 

I am not familiar with pulling practices yet. What is the next step for this PR?

kasperosterbye:
I am not sure, but now my feeling is that there is no acceptable solution - I believe that their tests and building is done using Linux and Mac, and not Windows. 
I have had similar issues where I place the change in overriding the simple method as an extension in some of my own code. 
It makes my code work. 
But if you think it is annoying, I agree.

Ducasse:
Hi guys 

Sorry that I missed the PRs. I'm on vacation and I will have a look.

Ducasse:
First I will do a release with the new Mustache version (1.4)

Ducasse:
I have a bug that is delaying the release so I integrated. Thanks for the contribution.

PR #1006, opened 2026-03-20
Feature/references

Fanirindrainy:


src/Microdown-Rules/MicReferencesChecker.class.st
```diff
@@ -0,0 +1,62 @@
+Class {
+	#name : 'MicReferencesChecker',
+	#superclass : 'MicChecker',
+	#category : 'Microdown-Rules-References',
+	#package : 'Microdown-Rules',
+	#tag : 'References'
+}
+
```
Ducasse, line 20: Could you add a class comment that explains what this checker is about

Fanirindrainy: Yes, I'll add class comment for this checker.

PR #1007, opened 2026-03-20
Feature/code indentation

Fanirindrainy:


src/Microdown-Rules/MicCodeIndentationChecker.class.st
```diff
@@ -0,0 +1,29 @@
+Class {
+	#name : 'MicCodeIndentationChecker',
+	#superclass : 'MicChecker',
+	#category : 'Microdown-Rules-CodeIndetation',
+	#package : 'Microdown-Rules',
+	#tag : 'CodeIndetation'
+}
```
Ducasse, line 11: Could you add a class comment that explains what this checker is about?

src/Microdown-Rules/MicCodeIndentationChecker.class.st
```diff
@@ -0,0 +1,29 @@
...
+{ #category : 'visiting' }
+MicCodeIndentationChecker >> visitCode: aMicCode [
+	| lines |
+	lines := aMicCode body lines.
+	lines do: [ :line |
+		line isEmpty ifFalse: [ 
+			"verify if the line begin with space"
+			(line first = Character space) ifTrue: [ 
+				"except the space in the code like nextPut: ' '"
+				(line trimLeft includesSubstring: 'nextPut: '' ''') ifFalse: [
```
Ducasse: I do not understand why there is a check with nextPut: 

I would imagine that we could look for Character space.

src/Microdown-Rules/MicCodeIndentationCheckerTest.class.st
```diff
@@ -0,0 +1,67 @@
...
+	#package : 'Microdown-Rules',
+	#tag : 'CodeIndetation'
+}
+
+{ #category : 'running' }
+MicCodeIndentationCheckerTest >> generateFilesystemExample [
+	| file |
+	"incorrect cases - space instead of tabs"
+	file := fileSystem workingDirectory / 'indentationWrong.md'.
+	file writeStreamDo: [ :steam |
```
Ducasse: steam -> stream

src/Microdown-Rules/MicCodeIndentationChecker.class.st
Fanirindrainy: Yes, I'll add class comment for this checker.

src/Microdown-Rules/MicCodeIndentationChecker.class.st
Fanirindrainy: You are absolutely right, the correct rule is simply to check wheter the line starts with a space; the content of the line should not matter. I'll correct this.

src/Microdown-Rules/MicCodeIndentationCheckerTest.class.st
Fanirindrainy: Oh, my bad, I'll correct this.

Ducasse:
I will merge it and we will be able to check on real books once this is integrated into Pillar. 
It will be a great validation

PR #1009, opened 2026-03-20
Feature/method structure

Fanirindrainy:


src/Microdown-Rules/MicMethodStructureChecker.class.st
```diff
@@ -0,0 +1,27 @@
+Class {
+	#name : 'MicMethodStructureChecker',
+	#superclass : 'MicChecker',
+	#category : 'Microdown-Rules-MethodStructure',
+	#package : 'Microdown-Rules',
+	#tag : 'MethodStructure'
+}
```
Ducasse, line 30: Can you add a comment because I do tno really understand what it is about?

Fanirindrainy: ok

PR #1011, opened 2026-03-20
Feature/method reference

Fanirindrainy:


src/Microdown-Rules/MicMethodReferenceChecker.class.st
```diff
@@ -0,0 +1,24 @@
+Class {
+	#name : 'MicMethodReferenceChecker',
+	#superclass : 'MicChecker',
+	#category : 'Microdown-Rules-MethodeReference',
+	#package : 'Microdown-Rules',
+	#tag : 'MethodeReference'
+}
```
Ducasse, line 23: Could you add a class comment that explains what this checker is about

src/Microdown-Rules/MicMethodReferenceChecker.class.st
```diff
@@ -0,0 +1,24 @@
...
+
```
Ducasse, line 24: What would be good is to be able to check that we get either Class>>#selector or Class>>selector.
because it depends on the book.

src/Microdown-Rules/MicMethodReferenceChecker.class.st
Fanirindrainy: Yes, I'll add class comment for this checker.

src/Microdown-Rules/MicMethodReferenceChecker.class.st
Fanirindrainy: That is exactly what our checker does. It detects when both forms coexist in the same doc; either always with # or always without #. If the document uses only one form consistently, no error is reported.

PR #1012, opened 2026-03-21
Feature/vocabulary

Fanirindrainy:


Ducasse:
Oh excellent it will be easier to plug checkers configurations this way.

src/Microdown-BookTester-Tests/MiCheckerEngineTest.class.st
```diff
@@ -0,0 +1,78 @@
...
+	| config results |
+	config := OrderedDictionary new.
+	results := engine runOn: file withConfiguration: config.
+	self assert: results isEmpty.
+]
+
+{ #category : 'running' }
+MiCheckerEngineTest >> testEngineWithUnknownChecker [
+	| config results |
+	config := OrderedDictionary new.
```
Ducasse, line 74: Yes this is good to check that wrong input in checker name.

src/Microdown-BookTester/MicCheckerEngine.class.st
```diff
@@ -0,0 +1,57 @@
...
+}
+
+{ #category : 'as yet unclassified' }
+MicCheckerEngine >> checkersFromConfiguration: aConfiguration [
+	| availableCheckers activeCheckers |
+	availableCheckers := Dictionary new.
+	MicChecker allSubclasses do: [ :aClass |
+		| instance |
+		instance := aClass new.
+		(instance respondsTo: #checkerName) ifTrue: [
```
Ducasse: We should make sure that all checkers inherit from MicChecker and that it implements on the class side the method checkerName. 
By default it can return 'UnamedChecker'
This way we can avoid the
```
instance := aClass new 
and respondsTo: 
```

In Pharo class methods are dynamically looked up :)

src/Microdown-Rules/MicVocabularyChecker.class.st
```diff
@@ -21,6 +21,21 @@ MicVocabularyChecker >> checkText: aMicText [
 							 patternPair: patternPair) ] ]
 ]
 
+{ #category : 'as yet unclassified' }
+MicVocabularyChecker >> checkerName [
+	^ 'Vocabulary'
+]
+
+{ #category : 'as yet unclassified' }
+MicVocabularyChecker >> configureFrom: aConfiguration [
```
Ducasse, line 30: May be should check that we get an collection and not a boolean

src/Microdown-BookTester/MicCheckerEngine.class.st
Fanirindrainy: ah ok :)

src/Microdown-Rules/MicVocabularyChecker.class.st
Fanirindrainy: Ok

Ducasse:
Same here I will integrate it so that after we can do another pass once pillar is using it and we get real feedback.

PR #1025, opened 2026-04-09
Feature/dashboard

Fanirindrainy:


src/Microdown-BookTester/MicAnalysisReportWriter.class.st
```diff
@@ -35,30 +34,24 @@ MicAnalysisReportWriter class >> checkersFromConfiguration: aConfiguration [
 					ifAbsent: [ Stdio stdout nextPutAll: 'The checker ', aCheck asString , ' is not available'].
 				configuredCheckers 
 					at: aCheck 
-					put: (availableCheckers at: aCheck) new.
-				self 
-					configureCheckers: configuredCheckers 
-					from: aConfiguration.
-				^ configuredCheckers values ]].
-		
+					put: (availableCheckers at: aCheck) new ].
```
Ducasse, line 37: This is correct I configured the checker far too many times.

src/Microdown-BookTester/MicChecker.class.st
```diff
@@ -59,6 +64,7 @@ MicChecker >> addResult: aResult [
 MicChecker >> checkProject: aFileReference [
 
 	| collector |
+	results := OrderedCollection new.
```
Ducasse, line 67: Can you explain why you need to reinitialize the results? 
Because results is already initiliazed in the initialize method.

Fanirindrainy: I found that In the normal usage through MicAnalysisReportWriter, a new checker instance is created for each check, so the reinitialization is indeed redundant. However, in the tests, the same checker instance is reused multiple times, which requires the reinitialization.

Ducasse: This is strange because the setUp is creating a new instance each time. 
BTW could you split the PR so that we can have a PR for the tools and one for the rest.

I plan also to make the list of checker settable. this way we will be able to write better tests.
Tx.

Ducasse: I fixed this. It was stupid :)

Ducasse:
Hi faniry 
thanks for the fix. Could you split the PR so that you send me all the UI/HTML because I modified the other parts and you will have conflict.

Ducasse:
To make it simpky you can fileout the classes you want and then in another image/branch you can drop the files on top of Pharo and it will load them. 
Else you can replay the changes you want from another image using the source in pharo-local/epicea (selecting repository and browsing the pharo-local/epicea files. This is quite handy.

Fanirindrainy:
Hi Stéphane,
I created a separate PR with only the UI/HTML part ([MicCheckerHTMLReporter.class.st]):
https://github.com/pillar-markup/Microdown/pull/1051

Thanks for the guidance!

Issue #1029, opened 2026-04-09
[LastPeriodInCodeChecker] Fix the explanation of the result so that it generates a list item


Ducasse:
and of course write a little test.

Issue #1028, opened 2026-04-09
codeEndingWithPeriodAndEmptyLines does not have an extra line


Ducasse:
```
file := fileSystem workingDirectory / 'codeEndingWithPeriodAndEmptyLines.md'.
	file writeStreamDo: [ :stream |
			stream nextPutAll: '

Authors may want to detect useless period as in the following code snippet

```
setX: xInt setY: yInt

	x := xInt.
	y := yInt.           
```
' ].
```

Ducasse:
You see this is what I meant.
We add a file with code with an empty line and we check.

Ducasse:
So apparently this was already working

```
testCheckerChecksCorrectlyMissingExtraPeriodWithExtraEmptyLines
	"may be the author wants all the code to finish with an extra period.
	so we configure the checker to do so."

	checker identifyMissingLastPeriod.
	checker checkProject: fileSystem / 'codeEndingWithPeriodAndEmptyLines.md'.
	self assert: checker isOkay
```

Ducasse:
But the files contained tab at the beginning of the lines so we will add a couple of tests with 
empty lines, 
empty lines with space
empty lines with tab.

Issue #1026, opened 2026-04-09
Write tests for Configuration combination only > ignored + configured


Ducasse:
We should have 
- one test that check that only the only checker is "executable"
- we can have one variation that checks that the checker is correctly configured
- one for ignored 
- one for ignored showing that others are still configured.
- may be can have one test showing that we can handle two checkers

Ducasse:
@antoine we should discuss because your tests are not testing this. 
The idea is that we should have a full configuration with ignored, only,....



```
{
	"language" : "en-UK",
	"base_url" : "",
	"attribution" : "K. De Hondt, S. Ducasse with S. Jordan Montaño and E. Lorenzano",
	"keywords" : "Spec, user interface, Pharo, builder, widgets",
	"title" : "Application Building with Spec 2.0",
	"series" : "Square Bracket Tutorials",
	"htmlWriter" : "html",
	"latexWriter" : "miclatex:sbabook",
	"site_name" : "Spec20",
	"tocFile": "index.md",
	"onlyCheckers" : [ "Vocabulary" ],
	"checkerConfigurations" : 
			[
			"Vocabulary" :
				[
					["sub-presenter", "subpresenter" ],
					["sub presenter","subpresenter"],
					["meta object","metaobject"]
				],
			"LastPeriodInCode" : "true"
				
	]
}
```

and we should check that 

- one test that check that only the only checker is "executable"
- as mentionned above.

Ducasse:
- [ x ] si configuration utilise une clef mauvaise
- [ x ] si configuration utilise une clef mauvaise interne dans un dictionary
- [ x ] si only seulement only est configure
- [ x ] multiple only
- [ x ] multiple are configured
- [ x ] si only ignored is not taken into account.
- [ x ] si ignored alors ignored, les autres sont configure
- si rien alors tous

Ducasse:
- Test that a checker with a bogus name does not break.

PR #1031, opened 2026-04-10
issue 1026/1028/1029

AntoineUsal:
issue #1026
issue #1028
issue #1029

add tests and correction files

### DIFF
--- a/src/Microdown-Rules/MicLastPeriodInCodeCheckerTest.class.st
+++ b/src/Microdown-Rules/MicLastPeriodInCodeCheckerTest.class.st
@@ -65,10 +65,11 @@ Authors may want to detect useless period as in the following code snippet
 setX: xInt setY: yInt
 
 	x := xInt.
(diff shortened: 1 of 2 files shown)


src/Microdown-Rules/MicLastPeriodInCodeCheckerTest.class.st
```diff
@@ -65,11 +65,11 @@ Authors may want to detect useless period as in the following code snippet
 setX: xInt setY: yInt
 
 	x := xInt.
-	y := yInt.
-     
+	y := yInt.        
+
```
Ducasse, line 69: To check that all the logic is working. 
We should add a specific virtula file with code block with an extra line and check that the checker validates in both cases. 
So I would expect that we have also to change the code of the checker (or may be it eats the empty line already.

Ducasse:
I will integrate it. 
Now it is better to do one PR per issue.

PR #1034, opened 2026-04-13
Add tests for ParagraphLabel

AntoineUsal:
#issue 1032

add tests


Ducasse:
Tx antoine. 
- Can you check the conflicts?
- Then could you use the isOk protocol as the other tests?
- I would be nice to get the explanation of the failures in the tests too.

PR #1049, opened 2026-04-17
Change true to false

### LINKED ISSUE #1048: The default for LastPeriodInCode should be false
Since in Pharo the code formatter removes the last period.
A period is not a terminator but separator.

AntoineUsal:
Fix #1048


Ducasse:
Do we have a test that cover it?

Ducasse:
We did it together so I will close this issue.

PR #1050, opened 2026-04-17
lastPeriodInLabelParagraph

### LINKED ISSUE #1033: lastPeriodInLabelParagraph
We should be able to define from the configuration the level of the paragraph. For example, from 4 to 6, (greater than 4 for example).
- Write tests that check that we can set the level for example 5 or above can be a paragraph and its period checked.
- Write tests for the configration.

AntoineUsal:
Fix #1033 

Add a new class test


Ducasse:
Why do we need a new class?
We already have LastPeriodInLabelParagraph?

Ducasse:
I will close this one because we already have this class.

PR #1051, opened 2026-04-18
Add htmlEscape and fix HTML reporter

Fanirindrainy:
Separate PR for UI/HTML only — contains only MicCheckerHTMLReporter with htmlEscape fix, as requested.


Ducasse:
TX!

PR #1053, opened 2026-04-20
- Rename the class and its test to follow the same naming convention …

Ducasse:
…than the other LastPeriod...


Ducasse:
I will close it and redo it once all the issue of Antoine are integrated to avoid creating conflicts.

PR #1056, opened 2026-04-21
Add test for issue 1048

AntoineUsal:
Fix issue #1048


AntoineUsal:
test crashed in an other class, but no changes

Ducasse:
Ok I will integrate it and do a pass on it.

Ducasse:
Antoine I'm checking and your changes were made on the wrong class. 
The level is for the paragraph.

Issue #1063, opened 2026-04-28
Urls reconsideration


Ducasse:
- check URLs case: when generating a website links are casesensitive and 
-- a checker that verifies that there is no file with the same name but different cases (on mac this is not possible) but on linux or windows it is
-- we can make sure that the HTML writer outputs lowercase files with lower case links.
 
- can we output files (and not url) with space. 
right now urls are encoded.

PR #1069, opened 2026-04-28
Fix error figure

AntoineUsal:
The PR is related to https://github.com/pillar-markup/Microdown/issues/1063
but I'm not sure that it is the correct answer.


Ducasse:
Can you provide a couple of tests?

Ducasse:
The main problem is that it does not work because the file still contain a space and not a %20. 
So may be we should flag this file as error.

PR #1080, opened 2026-05-13
Extend the reference checker

### LINKED ISSUE #1075: Extend the reference checker to spot reference to files that have a different case
- A checker that verifies that there is no file with the same name but different cases 
- Test with figures with different uppercase. 
- Tests for link resources with different cases

AntoineUsal:
Fix #1075

### DIFF
--- a/src/Microdown-BookTester-Tests/MicReferenceCheckerTest.class.st
+++ b/src/Microdown-BookTester-Tests/MicReferenceCheckerTest.class.st
@@ -923,3 +923,22 @@ MicReferenceCheckerTest >> testUnreferencedFigure [
 	self assert: unref anchorLabel equals: 'Fig1'.
 	self assert: unref sourceFileReference fullName equals: '/defFig1.md'
 ]
+
+{ #category : 'tests - single file' }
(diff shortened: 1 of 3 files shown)


src/Microdown-BookTester-Tests/MicReferenceCheckerTest.class.st
```diff
@@ -923,3 +923,22 @@ MicReferenceCheckerTest >> testUnreferencedFigure [
...
+	(dir / 'figures') ensureCreateDirectory.
+	(dir / 'figures' / 'MonImage.png') ensureCreateFile.
+	
+	file := dir / 'testWrongCase.md'.
+	file writeStreamDo: [ :stream | stream nextPutAll: '![Test image](figures/monimage.png)' ].
+	
+	visitor checkProject: file.
+	
+	self deny: visitor isOkay.
+	wrongCaseResults := (visitor results groupedBy: [ :each | each class name ]) at: #MicWrongCaseFileResult.
```
Ducasse, line 942: No need to group just to get a single result. 
I will integrate it and fix it.

PR #1088, opened 2026-05-21
Fix 2 tests from MicToCHTML

AntoineUsal:
Fix #1085


src/Microdown-HTMLExporter/MicHTMLVisitor.class.st
```diff
@@ -480,24 +480,23 @@ MicHTMLVisitor >> visitMonospace: aMonospace [
 ]
 
 { #category : 'visiting - list' }
```
Ducasse, line 482: Why the super invocation is not correct?
Then there is also visitChildrenOf: that is doing the same as the change proposed. 

If we take you change basically this is impossible to change the item rendering.

src/Microdown-Toc/MicHTMLVisitor.extension.st
```diff
@@ -8,19 +8,18 @@ MicHTMLVisitor >> visitTocEntry: aMicTocEntry [
 			canvas tag
 				name: 'li';
 				with: [
-					aMicTocEntry header headerElements do: [ :each |
-						each accept: self ] ] ]
+					aMicTocEntry header children do: [ :each | each accept: self ] ] ]
```
Ducasse: Why not keep the code as it is?
What is the difference because headerElements are the children

So this changes should have no impact

PR #1092, opened 2026-05-27
Fix checkCaption

AntoineUsal:
Fix #1090


src/Microdown-BookTester/MicFigureEmptyCaptionCheckerTest.class.st
```diff
@@ -99,6 +117,16 @@ MicFigureEmptyCaptionCheckerTest >> testConfigureFromCheckOnly [
 	self deny: checker checkFigure.
 ]
 
+{ #category : 'tests' }
+MicFigureEmptyCaptionCheckerTest >> testExplanationForConfiguration [
```
Ducasse, line 121: This is not enough we should have a real check. else 'a' can be a valid explanation.
Check the other tests for the configuration in the system to see how this is usually done

PR #1094, opened 2026-05-27
Fix block startLine

### LINKED ISSUE #1064: Add support for line numbers in the dom element
RIght now we cannot report the line of the element in the text and it would be good to be able to do it.

AntoineUsal:
Fix #1064

### DIFF
--- a/src/Microdown/MicParser.class.st
+++ b/src/Microdown/MicParser.class.st
@@ -268,18 +268,19 @@ MicParser >> handleLine: line [
 		-if the current block can consume the line, it manages it and this potentially creates a new block that becomes the current one.
 		When the line is not consumed, the current block is closed and its parent becomes the current one and the process is called back to treat the line."
 
-	| nextNode tempNode |
+	| nextNode |
 	(current canConsumeLine: (current massageLine: line))
 		ifTrue: [ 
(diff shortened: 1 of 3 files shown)


src/Microdown/MicParser.class.st
```diff
@@ -267,9 +268,16 @@ MicParser >> handleLine: line [
...
 
+	| nextNode tempNode |
 	(current canConsumeLine: (current massageLine: line))
 		ifTrue: [ 
-			current := current addLineAndReturnNextNode: line.
+			nextNode := current addLineAndReturnNextNode: line.
+			current == nextNode ifFalse: [ 
+				tempNode := nextNode.
+				[ tempNode notNil and: [ tempNode startLine isNil ] ] whileTrue: [ 
+					tempNode startLine: currentLineNumber.
```
Ducasse: I do not understand this logic why the startLine of a block element would be nil?
I would prefer that by construction when we create a new element we set the linenumber (the first line that created the element) 

I have the impression that handleLine: should only increase the line number and the line number should be set to a new bloc element at creation time (probably in the caller of handleLine:

Ducasse: on close we could also set a stop line this way we would know the size in lines of the elements.

PR #1100, opened 2026-06-01
Add new checker MicSecondaryRoot

AntoineUsal:
Fix #1099

### DIFF
--- a/src/Microdown-BookTester/MicChecker.class.st
+++ b/src/Microdown-BookTester/MicChecker.class.st
@@ -157,3 +157,9 @@ MicChecker >> rootDirectory: aFileReference [
 
 	rootDirectory := aFileReference
 ]
+
+{ #category : 'testing' }
+MicChecker >> visitSecondaryRootBlock: aSecondaryRootBlock [
+
+	^ self visitChildrenOf: aSecondaryRootBlock
+]
--- a/src/Microdown-BookTester/MicChecker.class.st
+++ b/src/Microdown-BookTester/MicChecker.class.st
@@ -157,9 +157,3 @@ MicChecker >> rootDirectory: aFileReference [
 
 	rootDirectory := aFileReference
 ]
-
-{ #category : 'testing' }
-MicChecker >> visitSecondaryRootBlock: aSecondaryRootBlock [
-
-	^ self visitChildrenOf: aSecondaryRootBlock
-]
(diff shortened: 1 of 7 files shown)


src/Microdown-BookTester/MicChecker.class.st
```diff
@@ -157,3 +157,9 @@ MicChecker >> rootDirectory: aFileReference [
 
 	rootDirectory := aFileReference
 ]
+
+{ #category : 'testing' }
+MicChecker >> visitSecondaryRootBlock: aSecondaryRootBlock [
```
Ducasse: We should add a visitSecondaryRootBlock: in the default MicrodownVisitor it can just call the  logic of rootblock eg. visitRootBlock:

src/Microdown/MicRootBlock.class.st
```diff
@@ -105,6 +105,11 @@ MicRootBlock >> initialize [
 	debugId := self class nextDebugId.
 ]
 
+{ #category : 'debugging' }
+MicRootBlock >> isSecondaryRoot [
```
Ducasse, line 109: It could be smarter. I wonder. Because a main root block parent is nil while by construction a secondaryRootBlock should be nested into something.

src/Microdown/MicSecondaryRootBlock.class.st
```diff
@@ -1,6 +1,9 @@
 Class {
 	#name : 'MicSecondaryRootBlock',
 	#superclass : 'MicRootBlock',
+	#instVars : [
+		'file'
```
Ducasse: If you look in the superclass fromFile is already defined so 
either we use fromFile implementation (I do not remember but may be it was with managed with a property and we could move it using an instance variable (and remove the implementation in secondaryRootBlock

src/Microdown/MicSecondaryRootBlock.class.st
```diff
@@ -12,6 +12,11 @@ MicSecondaryRootBlock >> accept: aVisitor [
 	^ aVisitor visitSecondaryRootBlock: self
 ]
 
+{ #category : 'testing' }
+MicSecondaryRootBlock >> fromFile: aFileOrResource [
+	self propertyAt: #file put: aFileOrResource
```
Ducasse, line 17: Je pense que nous avons la meme methode dans la superclasse

PR #1109, opened 2026-06-19
Fix setCurrent and add tests

AntoineUsal:
Fix #1108

### DIFF
--- a/src/Microdown-Tests/MicParserTest.class.st
+++ b/src/Microdown-Tests/MicParserTest.class.st
@@ -160,10 +160,10 @@ MicParserTest >> testSetCurrentDoesNotResetLineNumberForNormalBlock [
 	parser := MicParser new.
 	paragraph := MicParagraphBlock new.
 	
-	parser instVarNamed: 'currentLineNumber' put: 15.
+	parser currentLineNumber: 15.
 	parser setCurrent: paragraph.
 	
-	self assert: (parser instVarNamed: 'currentLineNumber') equals: 15.
+	self assert: parser currentLineNumber equals: 15.
 ]
 
 { #category : 'tests' }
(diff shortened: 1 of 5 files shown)


src/Microdown-Tests/MicParserTest.class.st
```diff
@@ -152,6 +152,30 @@ is really cool'.
...
+	self assert: (parser instVarNamed: 'currentLineNumber') equals: 15.
+]
+
+{ #category : 'tests' }
+MicParserTest >> testSetCurrentResetsLineNumberForSecondaryRoot [
+	| parser secondaryRoot |
+	parser := MicParser new.
+	secondaryRoot := MicSecondaryRootBlock new.
+	
+	parser instVarNamed: 'currentLineNumber' put: 15.
```
Ducasse: We should not reflective access to instanceVariable. 
Better define a setter and use it.

src/Microdown/MicParser.class.st
```diff
@@ -388,5 +388,6 @@ MicParser >> setCurrent: aBlock [
 	note for me: we could just inforce this. aBlock parent: current.
 	No this is not correct since 
 	the parent of an item list is the list and this is the list that is parent of environment."	
-	current := aBlock
+	current := aBlock.
+	aBlock isSecondaryRoot ifTrue: [ currentLineNumber := 1 ]
```
Ducasse: You do not need this test.
Remember Do not ask, tell

so here we should simply do 
aBlock incrementLineNumberOf: self 


And on 
```
RootBlokc >> incrementLineNumber: aParser
    aParser currentLineNumber := 1
```

```
MicElement >> incrementLineNumber: aParser
  ^ self 
```

